PR #1, opened 2026-07-31
Two test failures that were the environment, not the code

isenbek:
[CI](https://github.com/sqliteai/waste/actions/workflows/ci.yml)
<sub>status of upstream `main` — this PR's own run is still awaiting maintainer approval, as fork PRs do not start workflows on their own.</sub>

Three checks in `tests/run.sh` fail on any machine whose environment happens to differ from the author's, and both causes blame the code under test.

## The port the suite hardcoded

`tests/run.sh` used fixed ports 8731 and 8732 for the range server the download worker talks to. On a machine already running something on 8731, `range_server.py` cannot bind, the worker resumes against whatever else answers, and the failure surfaces as **"resume"** and **"state-file skip"** — two checks pointing at `tools/fetch_weights.sh`, which is fine. The third check passes only because 8732 happens to be free.

Nothing there needs a *particular* port, only a free one. `range_server.py` now takes port `0` and reports back what the kernel gave it through `--port-file`, written under a temp name and renamed so a caller polling for the file cannot read half a number. It is written after the constructor has bound and listened and before `serve_forever`, so a caller that sees the file can connect immediately — anything arriving early waits in the backlog rather than being refused.

That also removes the `sleep 1` before each server, which was the same bug from the other side: a fixed wait is either longer than it needs to be or, on a loaded machine, not long enough. `start_server` polls in 100 ms steps and stops early if the process died. When it cannot start one, the block says so and names the checks that did not run, instead of letting them fail as though they had.

## The package that shadowed `tests/`

The serve suite failed for an unrelated reason with the same shape: an unrelated `tests` package installed in site-packages shadows this directory, so `python3 -m unittest discover -s tests/serve -t .` cannot import `tests.serve` and reports the whole suite as one failure naming nothing.

`tests/` had no `__init__.py`, which makes it an implicit namespace portion, and the import system prefers a *regular* package found later on `sys.path` over an earlier portion. `tests/__init__.py` exists to lose that race deliberately, and says so in a comment so it does not read as an accident someone can delete.

## Verification

**24 passed, 0 failed, 11 skipped** in `tests/run.sh` on a clone with no container — the three failures above were the whole difference — and **167/167** in the serve suite with no `PYTHONNOUSERSITE` workaround. Verified with the conflicting daemon still holding 8731. `--no-range` confirmed still reaching the server (no `Accept-Ranges` in the response), and the CI SPDX guard still passes.

`CLAUDE.md` was split out into #2 — this PR is the test fix alone.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,11 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 SQLite Cloud, Inc.
+#
+# This file exists to make `tests` a regular package, and that is its whole
+# job. Without it the directory is only an implicit namespace portion, and
+# the import system prefers a *regular* package found later on sys.path —
+# so any unrelated `tests` package installed in site-packages shadows this
+# one and `python3 -m unittest discover -s tests/serve -t .` fails to
+# import tests.serve at all. That is not a hypothetical: it happened, and
+# the whole serve suite reported as one failure with no hint that the
+# module it could not find was ours.
--- a/tests/range_server.py
+++ b/tests/range_server.py
@@ -21,7 +21,27 @@ taken". --port-file is how the caller learns which port it got.
 import http.server
 import os
 import re
+import socketserver
 import sys
+
+
+class Server(http.server.HTTPServer):
+    """HTTPServer without the reverse DNS lookup in its constructor.
+
+    HTTPServer.server_bind calls socket.getfqdn() on the address it just
+    bound, and TCPServer.__init__ runs it *between* bind() and the listen()
+    in server_activate. For 127.0.0.1 the answer is worth nothing — nothing
+    here reads server_name — but on a macOS CI runner it blocked for about
+    fifteen seconds, and until it returns the port is bound and not yet
+    listening, so a caller connecting in that window is refused rather than
+    queued. It also delayed the port file below past the caller's patience:
+    the first server of a run failed to report in time, the second was fast
+    because the resolver had cached the answer by then.
+    """
+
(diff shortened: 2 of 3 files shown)


isenbek:
CI run for this branch's head (`451a216`), on my fork since fork PRs don't start workflows here without a maintainer's approval:

**https://github.com/isenbek/waste/actions/runs/30600971972** — all eight jobs green.

| job | |
|---|---|
| linux-x86_64, linux-arm64, macos-arm64 | ✅ `24 passed, 0 failed, 11 skipped` |
| macos-arm64 (Metal) | ✅ |
| asan + fuzz | ✅ |
| windows-x86_64 (cross-build), (run) | ✅ |

Worth recording that the first attempt at this **failed on macOS only**, and the second commit is the fix:

`HTTPServer.server_bind` calls `socket.getfqdn()` on the address it just bound, and `TCPServer.__init__` runs it between `bind()` and the `listen()` in `server_activate`. On the macOS runner that reverse lookup blocked ~15 s, so the port file appeared past the ten seconds `run.sh` waits — and only for the *first* server of the run, because the resolver had cached the answer by the second. Until it returns the port is bound but not listening, so anything connecting in that window is refused rather than queued, which is what the old `sleep 1` had been riding on.

Nothing here reads `server_name`, so the subclass skips the lookup: the download section now takes **0.36 s** on the macOS runner, against 35 s and a failure before. The same commit also kills a server that never reported — the failed job's log ended with `Terminate orphan process: pid (36234) (Python)`.

marcobambini:
Thanks @isenbek

Issue #7, opened 2026-07-31
The shipped oracle fixture does not match a container built by a default conversion


isenbek:
*engine → engine matches the PyTorch oracle* fails against a Kimi-Linear container converted with current defaults. The engine and the fixture agree on the answer and disagree on precision, which does not look like an engine error — but I cannot rule one out from the fixture alone, which is why this is an issue rather than a PR.

## Measurement

`tests/fixtures/oracle_kimilinear_16tok.bin` (655 360 bytes = 163 840 floats = vocab), against `test_forward` on the same 16 ids `tests/run.sh` uses:

```
lengths:      engine 163840   oracle 163840
argmax agree: True
correlation:  0.964251
max |diff|:   3.2840          mean |diff|: 0.5484      (check threshold: 1e-3)

engine top-5 ids [318, 5243, 387, 14509, 10484]
oracle top-5 ids [318, 5243, 14509, 387, 10484]
```

Same argmax, same top-5 set with two middle ranks swapped, correlation 0.96. That is the signature of the same computation over differently-quantized weights, not of a wrong one. A genuine engine bug would not track the reference this closely across 163 840 logits.

## Why I think it is the fixture

The fixture has not changed since `0d9c995` (`git diff --stat ce96e38..HEAD -- tests/fixtures/` is empty; md5 `e4b19db9c9abb503fcee46bc97cfe747`). `tools/kimi_ref.py` computes its logits **from a WASTE container**, so a fixture is only comparable to a container quantized the same way. My container is a plain `tools/convert.py --src ... --out ...` with no flags, giving a `Q4G/Q8G/F32` trunk.

The README's `3.6e-06` agreement is only achievable when the oracle and the engine read the same weights. Nothing records which conversion settings this fixture corresponds to, so a contributor converting from the released shards today gets a failure with no way to tell whether the engine or the fixture is at fault.

## What would settle it

Regenerating the oracle from the container under test with `tools/kimi_ref.py` and re-diffing. That is a pure-PyTorch 48 B forward pass on CPU, and I can run it and report the number if that is useful — it is the only version of this check that means anything on a locally-converted container.

## Possible resolutions, all yours to choose

- record the conversion settings the fixture was generated under, and have `tests/run.sh` skip with that reason when the container's trunk formats differ;
- ship the fixture for the default conversion instead, and regenerate it when the default moves;
- make the check generate its own oracle when `uv` and the source weights are present, and skip otherwise.

## Reproduced on

[...]

PR #8, opened 2026-07-31
The trunk width the load path could not read, and the container that hid it

### LINKED ISSUE #6: WASTE_Q8=0 cannot load a Q4G trunk, so the f32-equivalence check fails on any default conversion
`WASTE_Q8=0 ./test_forward CONTAINER ...` prints `load failed` on a Kimi-Linear container converted with current defaults, so *engine → int8 trunk storage == f32 weights* in `tests/run.sh` fails:

```
Traceback (most recent call last):
  FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp.o53zJCIBp3/f32.bin'
  FAIL  int8 storage changes results
```

The traceback is the symptom — the comparison script cannot find an output file, because the run that should have written it never loaded the model.

## Cause

`src/model.c:558`:

```c
} else if (!q8_off) {                             /* Q8G -> f32 */
    const int N = t->shape[t->ndim - 1];
    const int64_t rows = (int64_t)(t->n / (size_t)N);
    const int ng = (N + g - 1) / g;
    ...
    pread_all(fd, qbuf, (size_t)rows * ng * g, off)
```

The comment says `Q8G`, but the branch is `else if (!q8_off)` — it catches **every** non-F32 trunk format. It then reads `rows * ng * g` bytes, i.e. one byte per weight, which is only true for Q8G. `WQ_Q4G` is packed two per byte, so the read asks for twice what the tensor occupies, `pread_all` overruns, and `TRUNK_FAIL` fires.

## Why it shows up now

The trunk of a default `tools/convert.py` run is overwhelmingly Q4G — `--trunk-bits` defaults to 4. From the manifest of a container converted from `moonshotai/Kimi-Linear-48B-A3B-Instruct`:

| fmt | tensors | bytes/weight |
|---|---|---|
| 0 `WQ_F32` | 208 | 4.000 |
| 2 `WQ_Q8G` | 2 | 1.016 |
| 3 `WQ_Q4G` | **315** | **0.516** |

So every container produced by a default conversion has a trunk this path cannot dequantize.

**This is invisible to CI**, which is why the suite is green there: the synthetic container from `tools/make_test_container.py` has a `Q8G/F32` trunk (`waste info` confirms), so `WASTE_Q8=0` takes the one shape the branch handles. The check passes on the synthetic container and cannot pass on a real one.

## Which fix is right is your call

- teach the branch to unpack Q4G (and presumably Q3G, which has the same shape of problem), or
- restrict it to Q8G explicitly and have `tests/run.sh` skip when the trunk is not all-Q8, or
- treat `WASTE_Q8=0` as Q8-only by contract and say so where it is documented.

Happy to send a PR for whichever you prefer. I did not guess at it because the first option changes engine behaviour and the second changes what the suite claims to cover.

## Reproduced on

`v0.6.1` (`c4d45c5`), Linux x86_64 / AVX2, container converted from the released shards with `tools/convert.py --src ... --out ... --jobs 3` (no other flags). Same result on `0.6.0`.

isenbek:
Closes #6. Independent of #5 — this branch sits directly on `c4d45c5`.

## The bug

`WASTE_Q8=0` dequantizes the trunk once at load instead of at every use. Its branch was labelled `/* Q8G -> f32 */`, guarded `else if (!q8_off)`, and read `rows * ng * g` bytes — one byte per weight. True of Q8G alone, but the guard caught **every** non-F32 format, and `--trunk-bits` defaults to 4. On any container `tools/convert.py` produces today, the read asks for twice the bytes the tensor occupies.

What that did depended on what happened to follow the tensor in `trunk.bin`:

| container | behaviour |
|---|---|
| real (Kimi-Linear 49 B) | over-read runs off the end, `pread_all` short-reads, **load fails** |
| small | over-read succeeds into the **neighbouring tensor** — same prompt, different logits (`argmax 59` where the quantized path says `20`) |

The loud failure was the lucky case. Silently computing from the adjacent tensor's bytes is the one to worry about.

## The fix

The branch is gone. Non-F32 tensors are always loaded quantized, and when `!q8_off` the rows go through `waste_deq_row` — **the same unpacker the compute path calls** — into an f32 buffer, after which `q`/`qs` are freed and `bits`/`rowbytes`/`group` zeroed so the tensor is indistinguishable from one that was F32 on disk (`m->t` is calloc'd, and `unpack_row_i8` and friends branch on `bits` without consulting `q`).

Nothing in the load path knows a width any more, so nothing there can be wrong about one. Q3G is covered by construction rather than by a third copy of the unpacking — which is the shape of the original defect.

`embed_tokens` now stays on disk under `WASTE_Q8=0` as it does otherwise. Its rows go through the same unpacker when read, so the logits are unchanged and 1.11 GB stays out of the resident set.

## Why CI was green

`tools/make_test_container.py` wrote a **Q8G trunk throughout** — the one width the old branch handled. A real trunk is the other way round: `convert.py` writes Q8G for `embed_tokens` and `lm_head`, and Q4G for every layer weight (315 of 525 tensors on Kimi-Linear).

The synthetic container now does the same, so `waste info` reports the `Q4G/Q8G/F32` a real one does, and the check that had been passing for want of coverage now has some.

## Verification

- **Real container** (Kimi-Linear 49.12 B, converted from the released shards). `WASTE_Q8=0` loads where it used to fail, and its logits match the quantized path to **3.72e-05 max / 4.02e-06 mean** against a `1e-3` threshold. Both agree with a PyTorch oracle regenerated from that container (see #7) to ~5e-05. Suite **30 → 31 passed**.
- **Synthetic container**: `26 passed, 0 failed`. Reverting *only* `src/model.c` and keeping the new container gives `25 passed, 1 failed` — that is the regression coverage this was missing, and it now fails in CI rather than in a user's terminal.
- **ASan/UBSan** over the suite, now with the Q4G dequant inside it: `25 passed, 0 failed`, no diagnostics.
- **Fuzzer**, 200 malformed containers: `146 rejected, 54 still loaded, 0 crashed, 0 hung`.

Linux x86_64 / AVX2, `v0.6.1`.

The one remaining failure against the real container is #7 (the oracle fixture), which is unrelated and untouched by this.

### DIFF
--- a/src/model.c
+++ b/src/model.c
@@ -555,39 +555,7 @@ static int load_trunk(waste_model *m, const char *dir, const js_doc *d, int trun
             t->data = (float *)waste_dio_alloc(t->n * sizeof(float));
             if (!t->data) TRUNK_FAIL;
             if (pread_all(fd, t->data, t->n * sizeof(float), off)) TRUNK_FAIL;
-        } else if (!q8_off) {                             /* Q8G -> f32 */
-            const int N = t->shape[t->ndim - 1];
-            const int64_t rows = (int64_t)(t->n / (size_t)N);
-            const int ng = (N + g - 1) / g;
-            t->data = (float *)waste_dio_alloc(t->n * sizeof(float));
-            int8_t *qbuf = (int8_t *)malloc((size_t)rows * ng * g);
-            uint16_t *sbuf = (uint16_t *)malloc((size_t)rows * ng * sizeof(uint16_t));
-            if (!t->data || !qbuf || !sbuf ||
-                pread_all(fd, qbuf, (size_t)rows * ng * g, off) ||
-                pread_all(fd, sbuf, (size_t)rows * ng * sizeof(uint16_t), soff)) {
-                free(qbuf); free(sbuf); TRUNK_FAIL;
-            }
-            const int8_t *q = qbuf;
-            const uint16_t *sc = sbuf;
-            for (int64_t r = 0; r < rows; r++) {
-                for (int b = 0; b < ng; b++) {
-                    /* fp16 -> float without <arm_fp16.h> assumptions */
-                    const uint16_t h = sc[r * ng + b];
-                    const uint32_t sign = (uint32_t)(h >> 15) << 31;
-                    uint32_t exp = (h >> 10) & 0x1f, man = h & 0x3ff, bits;
-                    if (exp == 0) { bits = sign; }
-                    else { bits = sign | ((exp + 112u) << 23) | (man << 13); }
-                    float s;
-                    memcpy(&s, &bits, 4);
-                    for (int k = 0; k < g; k++) {
-                        const int col = b * g + k;
-                        if (col >= N) break;
-                        t->data[r * N + col] = (float)q[(r * ng + b) * g + k] * s;
-                    }
-                }
-            }
-            free(qbuf); free(sbuf);
-        } else {                             /* Q8G or Q4G, kept quantized */
+        } else {                             /* Q8G, Q4G or Q3G */
             const int N = t->shape[t->ndim - 1];
             const int64_t rows = (int64_t)(t->n / (size_t)N);
             const int ng = (N + g - 1) / g;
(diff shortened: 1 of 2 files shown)


isenbek:
Superseded by ae553de — closing.

`src/model.c` on `main` now loads every non-F32 trunk through `waste_deq_row` instead of the branch that assumed one byte per weight, which is what this PR proposed. The upstream fix also catches something this PR did not: the private dequantizer predated `waste_f16`'s subnormal fix and flushed group scales below 6.1e-05 to zero, so decoding through the shared unpacker fixes a second defect on the same lines.

`tools/make_test_container.py` likewise emits Q4G for layer weights now, so the regression has CI coverage — the gap that let this ship green.

Verified against a real container at ae553de (Kimi-Linear 49.12 B, converted from the released shards, Linux x86_64/AVX2):

- `WASTE_Q8=0` loads and generates where it previously failed
- `tests/run.sh` — `PASS quantized trunk storage == f32 weights`, the check this issue broke

#6 describes the defect and is fixed in `main`, though still open.

PR #17, opened 2026-08-03
docs: fix subject-verb agreement in docs/EFFICIENCY.md

lunar-me:
## Summary

Fixes a grammar issue in `docs/EFFICIENCY.md`.

## Problem

Subject-verb agreement error. 'all' is plural (referring to SpecMoEOff, SP-MoE, MoE-SpeQ), so the verb should be 'assume' not 'assumes'.

## Change

- **File**: `docs/EFFICIENCY.md` (line 312)
- **Before**: `all assumes`
- **After**: `all assume`

### DIFF
--- a/docs/EFFICIENCY.md
+++ b/docs/EFFICIENCY.md
@@ -309,7 +309,7 @@ env var — because the question recurs for every new container.
 ### D. Batching and speculative decoding — refuted for this machine
 
 The offloading literature — SpecMoEOff (2.5x), SP-MoE, MoE-SpeQ — all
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks for reading closely, but this one is correct as written.

The grammatical subject is *the offloading literature* — singular — and the em-dash list is an appositive, which does not change the number of the subject. Strip the parenthetical and the sentence reads "The offloading literature all assumes compute is free and the transfer is the wall", where *all* is a floating quantifier over a mass noun, as in "the evidence all points the same way". Agreeing the verb with the names inside the dashes is agreement attraction rather than a fix.

PR #18, opened 2026-08-03
docs: fix subject-verb agreement in docs/LEARNED.md

lunar-me:
## Summary

Fixes a grammar issue in `docs/LEARNED.md`.

## Problem

Subject-verb agreement. 'glibc/gcc' refers to two separate components (the glibc library and the gcc compiler), requiring the plural verb 'do not'.

## Change

- **File**: `docs/LEARNED.md` (line 602)
- **Before**: `glibc/gcc does not`
- **After**: `glibc/gcc do not`

### DIFF
--- a/docs/LEARNED.md
+++ b/docs/LEARNED.md
@@ -599,7 +599,7 @@ first time GitHub ran the workflow rather than by anyone reading the code.
 **`test_image` never linked on Linux.** Its rule passed `$(LDFLAGS)` —
 empty — where every other rule passes `$(LDLIBS)`, so the binary linked
 without `-lm`. It worked on macOS for a week because clang folds the one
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks, but `glibc/gcc` here is one thing — the GNU toolchain, named by the two parts that matter for this defect — not two subjects. A slash compound of that kind takes singular agreement, and the clause is parallel to "clang folds the one `sqrt()` … into an instruction" immediately before it. Had two actors been meant, it would say "and".

One thing worth knowing for future PRs: `docs/LEARNED.md` is append-only and dated (see CLAUDE.md). Its entries are records of what was found on a given day, so they are not retouched after the fact.

PR #19, opened 2026-08-03
docs: fix incorrect past tense 'costed' in docs/GATES.md

lunar-me:
## Summary

Fixes a grammar issue in `docs/GATES.md`.

## Problem

"costed" is not the correct past tense of "cost" in this context. The past tense of "cost" (meaning "had a cost of") is "cost", not "costed". "Costed" means "priced" or "estimated the cost of" in specific contexts, which is not the intended meaning here.

## Change

- **File**: `docs/GATES.md` (line 132)
- **Before**: `it costed only the I/O`
- **After**: `it cost only the I/O`

### DIFF
--- a/docs/GATES.md
+++ b/docs/GATES.md
@@ -129,7 +129,7 @@ simulation above it. The 1.5 tok/s projection for K3 on 64 GB stands.
 > floor below is the single most predictive number in the project — but
 > the projection assumed a cache several times larger than a 64 GB
 > machine can actually give K3 once the 27.28 GB trunk is resident, and
-> it costed only the I/O. Measured sweep in [LEARNED.md](LEARNED.md) §12.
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks, but *costed* is deliberate here, and the change would invert the meaning.

It is the transitive accounting sense — to cost something is to put a price on it ("the plan was costed at £2m") — not the past tense of "to cost £2m". The claim is that the 1.5 tok/s projection **costed only the I/O**: it charged for disk time and nothing else. That is the entire point of the note, and the longer version of the same finding further down the same file spells it out — "And it counts disk time only, where expert I/O is 53.5% of a cold decode step, so even instantaneous reads would less than double it."

"it cost only the I/O" would instead say the projection was cheap, which is not a claim about anything.

PR #20, opened 2026-08-04
README.md now offer the converted K3 container over BitTorrent

pierre-x:
Offer the converted K3 container over BitTorrent, ahead of converting it.

Converting from the published weights is the only route README documented, and it is the expensive one: a 1.42 TB source download, 4.7 hours of conversion, and staging storage that has to exist somewhere before it can be freed. The output is a 982 GB directory that is byte-identical for everyone who runs the default conversion, so nearly all of that cost is paid to reproduce a fixed artifact. This new magnet link skips it. The only downside is that a new torrent will have to be created if the WASTE weights format changes in a future release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,9 +105,34 @@ make check
 
 `make` builds the `waste` CLI and `libwaste.a`. `make check` creates a small synthetic model, so it does not download weights.
 
-### Get Kimi K3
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks a lot @pierre-x , much appreciated!

PR #26, opened 2026-08-05
Read DeepSeek-V3 checkpoints in the converter

fab2s:
# Read DeepSeek-V3 checkpoints in the converter

Two commits, both needed before `convert.py` can produce a container from a
DeepSeek-V3 family checkpoint (V3, R1, Kimi-K2). Independent of #27, which
fixes a separate defect in the engine's attention path — that one builds and tests
without this, but reproducing it needs a container, so this comes first.

**fp8 block scales.** These checkpoints store weights as `F8_E4M3` with a per-tile
`_scale_inv` companion. Both safetensors readers now apply it — `mxfp4.ST` and
`convert.py`'s own `ShardReader`, which is a second reader and was easy to miss.
The tile size is read from `quantization_config.weight_block_size` rather than
inferred from the weight and scale shapes: inferring looks possible and is wrong
whenever a dimension is not a multiple of the tile, and a compatible-but-wrong
size passes the shape check while placing every scale on the wrong rows.

**MoE tensor names.** Mixtral and Kimi use
`block_sparse_moe.experts.E.w1/w3/w2`; DeepSeek uses
`mlp.experts.E.gate_proj/up_proj/down_proj`. The layout is detected from what is
on disk and normalised to one spelling, so the engine still sees one name. Without
it the expert probe misses on every layer and a run reports `0 MB [missing]` 60
times after the download has already completed.

**MoE config keys.** The same split one level up, and this half fails differently:
`src/model.c` reads `num_experts`, a DeepSeek config only spells it
`n_routed_experts`, so `cfg_from_json` yields 0 experts and the bank loop in
`waste_model_load` refuses the finished container with no diagnostic. Also
`num_experts_per_tok` against `..._per_token`, which is one letter and leaves
`top_k` at 0. Normalised into the manifest where the tensor names already are, and
written only when absent so a config using the canonical spelling wins.

`moe_renormalize` is handled separately because `model.c` keys it on the field
being **present** rather than on its value — a plain alias of `norm_topk_prob`
would enable renormalisation for a checkpoint that sets it false. Emitted only
when true.

## Testing

`make check` with `WASTE_REF_MODEL` pointed at a default VQ3R Kimi-Linear
container. `tools/../test_fp8_blocks.py` covers the block mapping on synthetic
tensors (aligned tiles, partial tiles on both axes, a missing companion, a gross
size mismatch) and cross-checks one real tensor bit-exactly against an
index-array dequant; it also documents the case that cannot be caught, a
compatible-but-wrong tile size, which is why the value is read from config.

The `tests/test_convert_resume.py` change is in the first commit rather than its
own: that file stubs `mxfp4` with a `types.ModuleType`, so the new import has to
resolve there or every check in the file fails at module load, and the commit
would not be green on its own.

Verified end to end by converting `Kimi-K2-Instruct` — 61 layers, 384 experts
top-8, VQ3R, 354 GB expert set, 6.9 GB trunk — and opening it with `waste info`:
1.03 T total, 31.69 B active per token.

Nothing outside `tools/` and that one test file changes. `docs/LEARNED.md`,
`CHANGELOG.md` and `WASTE_VERSION_*` are deliberately untouched.

### DIFF
--- a/tests/test_convert_resume.py
+++ b/tests/test_convert_resume.py
@@ -97,6 +97,10 @@ def install_stubs():
         def tensor(self, name): return FakeTensor()
         def names(self): return [TRUNK_TENSOR]
     mx.ST = ST
+    # convert.py imports this alongside ST for fp8 block-scaled checkpoints.
+    # Resume never reaches a dequant, but the stub has to satisfy the import
+    # or every check in this file fails at module load.
+    mx.unblock_scale = lambda q, scale, block: q
     sys.modules["mxfp4"] = mx
     sys.path.insert(0, os.path.join(REPO, "tools"))
     import convert
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -86,7 +86,7 @@ def atomic_text(path, value):
     os.replace(tmp, path)
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
-from mxfp4 import ST                                            # noqa: E402
+from mxfp4 import ST, unblock_scale                             # noqa: E402
 
 # --- native VQ encoder (optional; ~15x the torch path) --------------------
 _VQ = None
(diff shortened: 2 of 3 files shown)


marcobambini:
Merged.

Verified before landing: CI 9/9, clean merge onto `main`, and locally
`tests/test_convert_resume.py` passes with the `mxfp4.unblock_scale` stub in
place — putting that stub in the first commit rather than its own was correct,
the file would not have been green otherwise.

The check that mattered for the models this repo already ships is that none of
this reaches them. `normalise_cfg` run against an existing Kimi-Linear
manifest's `config` returns an identical dict, no keys added. `moe_layout`
probes the Mixtral spelling first, so K3 and Kimi-Linear take the branch they
always did, and `trunk_rename` is the identity unless `.mlp.experts.` is in the
index. Purely additive, which is what makes it safe to land ahead of a release.

The reasoning I want on the record is the tile size. Reading it from
`quantization_config.weight_block_size` instead of inferring it from the two
shapes is right for exactly the reason you give: 300 rows against 3 scale rows
admits both 128 and 100, both pass the shape check, and only one is true. That
is a silent-wrong-answer failure — it would have surfaced downstream as VQ
reconstruction error nobody could attribute, on a conversion that takes hours.
The same goes for making a missing `_scale_inv` an error rather than a
fallback.

**One thing is missing.** The PR body describes `test_fp8_blocks.py` — aligned
tiles, partial tiles on both axes, a missing companion, a gross size mismatch,
and a bit-exact cross-check of one real tensor against an index-array dequant —
and it is not in the diff. That is the only test covering the path with the
silent failure mode, and by your own account it is already written. A follow-up
PR putting it under `tests/` is enough; nothing else is outstanding.

Landed before #27 so that the reproduction path for the rotary fix is available
from a single checkout. Both are in the 0.6.8 section of `CHANGELOG.md`.

marcobambini:
Follow-up on the one outstanding item from the merge review: the `test_fp8_blocks.py` your description covers is now tracked as #39, so it does not sit in a closed PR.

Nothing new is being asked there beyond what you already described. It adds two suggestions — the both-axes-partial corner, where the two independent `repeat_interleave` ceilings and the `[:M, :N]` crop meet, and a check that `ST` and `ShardReader` agree, since each parses `weight_block_size` separately and your own description is what pointed out that the second reader was easy to miss.

@fab2s no obligation — if you would rather not, say so on #39 and I will write it from your description.

PR #28, opened 2026-08-06
Correct torrent magnet in the README

pierre-x:
The torrent was generated with mktorrent using its default piece length of 256 Kb. That default is fine for a few GB; it is not fine for a TB payload: the torrent stores one 20-byte SHA-1 per piece, so 4,018,567 pieces produce ~80 MB of hashes in the info dictionary alone. That breaks libtorrent-based clients.
So I regenerated another one with 16 MiB pieces.

This eases, but does not answer, sqliteai/waste#24: it is still a torrent, and the ask there is an IPFS or plain-HTTP mirror for users who cannot reach BitTorrent at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ disk. Any BitTorrent client will do; the torrent's own piece hashes verify the
 container as it arrives.
 
 ```text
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks a lot @pierre-x

Issue #14, opened 2026-08-02
Linux auto-budget can overestimate usable memory under host pressure or cgroup limits


mccoyspace:
## Problem

When `ram_budget_bytes == 0`, WASTE currently caps its automatic choice at 7/8 of installed physical RAM. On Linux, that can be larger than the memory the process can safely allocate when the host already has resident workloads or a cgroup-v2 ancestor imposes a lower limit.

This is the Linux/open-time counterpart to the paging cliff already documented in [docs/LEARNED.md §16](https://github.com/sqliteai/waste/blob/main/docs/LEARNED.md#16-too-much-cache-is-worse-than-too-little-2026-07-28): the physical cap remains useful, but it is not sufficient when current host pressure or a finite cgroup is tighter.

This came out of the GN100/K3 investigation, with one important correction: the original 80.64 GiB paging sample was caused by a second WASTE model open after budget selection. Clean 80.64 GiB replays were safe. This proposal protects against pressure already present when `waste_open` begins; it cannot predict a workload that starts later.

## Proposed policy

For automatic budgets only, bound the ceiling by the minimum of:

- the existing 7/8 physical-RAM cap;
- Linux `MemAvailable` minus a 1/8 host-RAM reserve; and
- finite cgroup-v2 ancestor headroom (`memory.max - memory.current`) minus 1/8 of that group effective capacity.

The cgroup reserve is based on `min(memory.max, physical RAM)`, rather than host RAM alone. Otherwise an 8 GiB group on a 128 GiB host incorrectly loses a 16 GiB reserve. The reader walks ancestors so an unlimited leaf cannot hide a finite parent.

If a known live/cgroup ceiling cannot hold the model floor, the candidate fails before model-sized allocation. This deliberately differs from the current behavior of running at the floor and warning, so I would value guidance on that choice.

Explicit nonzero budgets remain caller policy: they are not silently clamped, but receive a warning above the current safe ceiling. Missing or malformed Linux telemetry falls back safely, non-Linux behavior remains unchanged, and cgroup v1 is deliberately out of scope.

## Tested candidate

I have this isolated against current `main` in [candidate commit 5e4147f](https://github.com/mccoyspace/waste-spark/commit/5e4147f1ebc79e04ed31df70beccb24fafb5b48e).

Validation completed:

[...]

Issue #11, opened 2026-08-01
CUDA backend: what would have to be true for it to pay


marcobambini:
`WASTE_ENABLE_CUDA=1` has been a reachable Makefile flag since the skeleton, and there is still no `src/cuda.cu` behind it — the build stops with a message saying so, which is the honest state but not a decision. This issue is to record what the decision would have to rest on, because "write the kernels" is not the work.

## Why this is not a backend

Metal is the one accelerator that got written, and it is **off by default because it was measured slower**:

| | total | | |
|---|---|---|---|
| CPU/NEON | 15.93 s | | |
| Metal | 19.43 s | | **22% slower** |

`lm_head` is the clearest case, one dispatch per token over 1.17 GB of int8 weights: **CPU 6 ms against GPU 22 ms** — 195 GB/s against 53 GB/s. The CPU path was already running at the machine's memory bandwidth on a bandwidth-bound matvec, so there was no headroom to take, and the GPU paid a synchronous round-trip per call on top.

The conclusion in `docs/BACKENDS.md` is about shape rather than about Metal, and it applies unchanged to CUDA:

> This engine issues several hundred small dependent matvecs per token and waits for each; that is the worst possible fit for an accelerator. Making the GPU pay would mean moving the whole forward pass on-device so there is one dispatch per layer instead of six, with the residual stream never returning to the host — **a different engine, not a backend.**

So the CUDA slot cannot be filled the way the AVX2 and AVX-512 slots were. Overwriting a few function pointers with kernels reproduces the Metal result on different hardware.

## What is genuinely different about a discrete NVIDIA card

One clause of that argument does not transfer, and it is the reason this is worth reopening rather than closing.

Metal was measured on Apple silicon with unified memory, where "the CPU is already at the machine's bandwidth" and "the GPU is at the machine's bandwidth" are the same sentence. A discrete card has its own memory at roughly an order of magnitude more bandwidth than a desktop CPU's. **That headroom is a vendor specification, not something measured here** — this repo has no NVIDIA hardware, which is also why `docs/BACKENDS.md` records CUDA as untested and untestable.

What does not change is the round-trip. The headroom only becomes reachable if the residual stream stays on-device, which is the same "different engine" the Metal work identified.

## The VRAM the design implies

If the forward pass moves on-device, the floor is whatever has to stay resident there. From `waste plan` on the containers here:

| | trunk resident | working set / token | container on disk |
|---|---|---|---|
[...]

PR #29, opened 2026-08-07
Add opt-in exclusive model-container ownership

mccoyspace:
## Problem

A dedicated WASTE host may want one process to own a large model container so
that a second cooperating launch is refused before model-sized allocation.
This is host policy rather than a data-safety or RAM-accounting invariant:
containers are read-only, and container identity does not describe total
machine memory pressure.

## Change

- Add opt-in `waste_cfg.exclusive_open` and `--exclusive-open` flags for the
  CLI and server; ordinary concurrent opens remain the default.
- Take a non-blocking advisory POSIX `flock` on the container directory before
  planning or model-sized allocation when exclusivity is requested.
- Share a device/inode-keyed, reference-counted ownership entry between
  contexts in the same process.
- Return `WASTE_E_BUSY` only for actual `EWOULDBLOCK`/`EAGAIN` contention.
  Directory-open, `fstat`, unsupported-`flock`, and other locking failures
  proceed without ownership, leaving model readability to the normal loader.
- Release ownership on the last close and on planning, budget, and partial-load
  failures; use close-on-exec and clear inherited registry state after fork.
- Leave Windows lifecycle behavior unchanged.

The lock is advisory between cooperating WASTE processes. It is not RAM
accounting, a filesystem lease, or a security boundary.

## Correctness

The focused process test covers:

- default concurrent behavior and opt-in exclusivity;
- same-process reference sharing and last-close release;
- competing `exec`, default bypass, and fork handling;
- search-only-directory fail-open behavior;
- under-budget, malformed-plan, and partial-load cleanup paths; and
- close-on-exec behavior.

Current upstream-base validation on macOS:

- `tests/run.sh /nonexistent`: 31 passed, 0 failed, 13 skipped;
- server suite: 168 checks;
- manual CLI and server contention diagnostics; and
- default CLI open while an independent process holds the directory lock.

The lock is acquired once per context lifetime and is absent from the token
path, so no throughput claim is made.

## Compatibility

There is no container-format, arithmetic, routing, state, or I/O change.
Concurrent opens remain the default. Version 0.6.7 and its CHANGELOG entry
record the appended `waste_cfg` field and new status; pre-1.0 binary clients
must rebuild, matching the existing ABI policy.

## Rollback

Hosts that do not pass `--exclusive-open` retain the prior behavior. Reverting
this one commit removes the optional ownership mechanism and public status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,41 @@ measurement is the useful part.
 `docs/LEARNED.md` carries the full reasoning; this file carries what
 changed. Each entry names the section to read for the numbers behind it.
 
+## 0.6.7 — 2026-08-10
+
+The container format, arithmetic, routing, and default open behavior are
+unchanged. A host can now ask cooperating WASTE processes for exclusive
+ownership of one container before model-sized allocation begins. This is a
+host-policy mechanism for a workstation daemon, not RAM accounting or a data
+lock: containers remain read-only, concurrent opens remain the default, and an
+unsupported advisory lock does not make a readable model unavailable.
+
+**Callers must recompile against this header.** `exclusive_open` was appended
+to `waste_cfg`, and `WASTE_E_BUSY` was added to `waste_status`. The library is
+still pre-1.0 and does not promise a stable ABI; `serve/engine.py`'s ctypes
+mirror moved with the C header.
+
+### Added
+
+- **Opt-in single-process container ownership**
+  ([#29](https://github.com/sqliteai/waste/pull/29)). On POSIX hosts,
+  `waste_cfg.exclusive_open`, or `--exclusive-open` in the CLI and server,
+  takes a non-blocking advisory `flock` on the container directory. Multiple
+  contexts in one process share a device/inode-keyed reference; a cooperating
+  process that also requests exclusivity receives `WASTE_E_BUSY`. The last
+  close and every planning, budget, and partial-load failure release ownership;
+  descriptors are close-on-exec, and a forked child discards the copied
+  registry. Windows keeps its existing lifecycle behavior.
+
+### Fixed
+
+- **Advisory locking fails open when ownership cannot be established.** Only
+  actual `EWOULDBLOCK`/`EAGAIN` contention returns `WASTE_E_BUSY`. A directory
+  that is search-only, a filesystem without `flock`, or another non-contention
+  locking failure continues through the ordinary model-open path. This keeps
+  external FUSE, SMB, and NFS containers usable and leaves their real read
+  errors to the existing loader diagnostics.
+
 ## 0.6.6 — 2026-08-05
 
 The engine decodes exactly as 0.6.5 did and no container format moved. Two
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ waste$(EXE): cli/main.o libwaste.a
 # the two failures tests/run.sh was written to catch, so a binary that
 # `test` builds and `clean` forgets defeats the check meant to notice it.
 TESTNAMES := test_kda test_container test_forward test_tokenizer test_k3parts \
-             test_state test_vision test_image test_memory test_cpus sweep
+             test_state test_vision test_image test_memory test_cpus test_lock sweep
 TESTBINS  := $(addsuffix $(EXE),$(TESTNAMES))
 
 test: $(TESTBINS)
@@ -272,6 +272,9 @@ test_memory$(EXE): tests/test_memory.o src/memory.o
 # it can check placement without a container, and the parse half of it runs
 # on the platforms that cannot bind a thread.
 test_cpus$(EXE): tests/test_cpus.o
+	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
+
+test_lock$(EXE): tests/test_lock.o libwaste.a
 	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
 
 %.o: %.c
--- a/cli/main.c
+++ b/cli/main.c
@@ -112,7 +112,7 @@ typedef struct {
     uint64_t budget;
     uint32_t ctx, max_tokens;
     float temperature, top_p;
-    int top_k, threads, quiet, learn, json, no_echo;
+    int top_k, threads, quiet, learn, json, no_echo, exclusive_open;
     int media_inlined;              /* the media block is already in the
                                        prompt string, inside the user turn */
     uint64_t seed;
@@ -190,6 +190,7 @@ static int parse_opts(int argc, char **argv, int from, opts *o)
         else if (!strcmp(a, "--json")) o->json = 1;
         else if (!strcmp(a, "--raw")) o->raw = 1;
         else if (!strcmp(a, "--verify")) o->verify = 1;
+        else if (!strcmp(a, "--exclusive-open")) o->exclusive_open = 1;
         else if (!strcmp(a, "-")) {                /* explicit stdin */
             if (o->n_pos >= MAX_POS) { fprintf(stderr, "too many arguments\n"); return -1; }
             o->pos[o->n_pos++] = "-";
@@ -273,6 +274,7 @@ static waste_status open_model(const char *path, const opts *o, waste_ctx **ctx)
      * Kimi-Linear. Worth it for a container that was copied or downloaded
      * and has not been read since. */
     cfg.verify_records = o->verify;
+    cfg.exclusive_open = o->exclusive_open;
     const waste_status st = waste_open(path, &cfg, ctx);
     /* Two statuses that say nothing useful on their own when --cpus is
      * what produced them, and it usually is: nothing else here can be
@@ -283,6 +285,10 @@ static waste_status open_model(const char *path, const opts *o, waste_ctx **ctx)
     else if (o->cpus && st == WASTE_E_UNSUPPORTED)
         fprintf(stderr, "--cpus: this platform does not bind threads to "
                         "CPUs (Linux and Windows only)\n");
+    else if (o->exclusive_open && st == WASTE_E_BUSY)
+        fprintf(stderr, "--exclusive-open: another process owns this "
+                        "container; stop it or retry without "
+                        "--exclusive-open\n");
     return st;
 }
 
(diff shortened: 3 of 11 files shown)


marcobambini (changes requested):
Thanks for this — the implementation is careful, and the test is the best
part of the PR: verifying ownership with an independent `flock` rather than
trusting the library's own return, and using `exec` instead of relying on
fork's copied registry, is exactly the right paranoia. I built the branch and
ran it: `tests/run.sh /nonexistent` gives 37 passed, 0 failed, 9 skipped on an
M-series macOS host, and `test_lock` passes.

Two things need to change before this can land: one defect, and the polarity
of the default.

## The lock turns any locking failure into an open failure

In `model_lock_acquire`, only `EWOULDBLOCK`/`EAGAIN` becomes `WASTE_E_BUSY`.
A failing `open()` on the directory, a failing `fstat()`, and every other
`flock` errno become `WASTE_E_IO`, and the container cannot be opened at all:

```
$ chmod 0111 tiny.waste          # directory traversable but not readable
$ ./waste info tiny.waste
open: I/O error
$ ./waste info tiny.waste --allow-concurrent-open
WASTE 0.6.6 (container v0, backend NEON, crc32 armv8, arm64)   # and on main
```

A search-only model directory is a legitimate way to publish a shared
read-only tree, and it is the mild version of the problem. The serious one is
that `flock` is absent or returns `ENOTSUP`/`EOPNOTSUPP`/`ENOLCK` on several
FUSE mounts (third-party exFAT and NTFS drivers on macOS), on SMB, and on
some NFS configurations — which is precisely the kind of external disk this
project puts a 982 GB container on. The user gets `I/O error` and nothing
that points at the new flag.

The fix is small: treat contention as `EWOULDBLOCK`/`EAGAIN` **only**, and let
every other failure proceed without ownership instead of failing the open.
Whether the container is actually readable is still decided two lines later by
`waste_plan_memory`, with the same `WASTE_E_IO` that has always meant that.

## The default should be inverted: opt-in exclusivity, not opt-out

The mechanism is fine. Having it on by default is the part I do not think
survives contact with how the engine is used.

**It guards the wrong identity.** The hazard you describe is RAM
oversubscription, and container identity is neither necessary nor sufficient
for it. Two processes opening *different* containers oversubscribe exactly as
badly and are untouched by this. Two processes opening a *small* container —
Kimi-Linear on a 128 GB workstation — do not oversubscribe at all, and are now
refused. The predicate that actually predicts the harm is resident bytes
against `waste_usable_ram()`, which `waste_open` already computes; the
container's inode is a proxy that is wrong in both directions.

**It cannot hold the invariant it claims, so it should not be a default.**
The lock is advisory and only binds cooperating WASTE processes. Any other
memory-hungry process on the machine, any WASTE process that passed the
opt-out, any second container — all unprotected. A default that stops the
tidy case and misses the untidy ones buys less than it costs.

**There is no data hazard to justify it.** A container is read-only, opened
`O_DIRECT`, never written by the engine. Nothing about two readers is unsafe;
the only cost is performance, on a machine the operator owns and can observe.
Refusing an operation the OS would happily allow, on an artifact where
concurrency is harmless, is a strong claim for a library to make on its host's
behalf — and `CLAUDE.md`'s library-first rule puts exactly this class of
decision (like logging, signal handling and argument parsing) with the host.

**The failure modes are asymmetric, and this inverts them the wrong way.**
Today, a mistaken second load is *visible and recoverable*: throughput drops,
the operator notices, kills one. After this change, a legitimate second load
is *invisible and blocking*: a workflow that worked yesterday now stops with a
message about a flag nobody has heard of, and on a filesystem without `flock`
support it stops for a reason unrelated to concurrency at all. Trading a
self-announcing performance problem for a silent availability problem is a bad
trade even when the performance problem is real.

**It breaks ordinary single-machine workflows.** With the default on,
`waste info MODEL`, `waste bench MODEL`, and a second `python3 -m serve MODEL`
all fail while a server is up (verified: `open: container is already open in
another process`). Anyone with one machine and one container hits this on the
first day — including inspecting the container that is currently being served,
which is a read-only question that never needed a model load to be exclusive.

**And the polarity is a one-way door.** Shipping opt-in and promoting it to
default later is a normal deprecation. Shipping default-on and retreating is a
behavioral break for every host that had been opening the same container more
than once.

Concretely, I would keep everything in this PR and change one thing: rename the
field to `waste_cfg.exclusive_open` with the sense reversed, and the flags to
`--exclusive-open` on the CLI and the server. Same code, same registry, same
refcounting, same test — `test_lock` only needs its `allow` arguments flipped.
Hosts that genuinely want single-owner semantics (a workstation daemon that
owns its model) set one field and get the behavior you built; everybody else
keeps working. If you want the RAM-pressure story addressed on its own terms
afterwards, the place for it is the budget arithmetic in `waste_open`, not the
container's inode — that would catch the different-container case this misses,
and it is worth a separate PR.

## Also needed

- **Version bump and a CHANGELOG entry.** `allow_concurrent_open` is appended
  to `waste_cfg` and there is a new enumerator, so the ABI moved. 0.6.6 set the
  precedent for `cpu_list` ("Callers must recompile against this header"), and
  per `CLAUDE.md` the changelog is updated in the same commit that bumps
  `WASTE_VERSION_*`. This case is worse than `cpu_list`: a binding built
  against 0.6.6's struct passes four bytes short, and `waste_open` reads the
  new field from past the end of the caller's object — garbage that happens to
  be non-zero silently disables the lock, so the misread fails quiet and in
  the direction that defeats the feature. The `serve/engine.py` mirror is
  updated correctly.

- **`tests/run.sh` discards the diagnostics.** `./test_lock ... 2>/dev/null`
  throws away exactly the `FAIL line N: ...` lines the test emits, so a CI
  failure is one line with no assertion behind it. The `else` branch also
  reports FAIL when `make_test_container.py` is what failed — a missing
  prerequisite, which this suite reports as SKIP by convention.

- **Nothing suggests the opt-out on `WASTE_E_BUSY`.** Both `cli/main.c` and
  `serve/__main__.py` already have the block that turns an opaque status into
  advice (the `--cpus` pair); this is the case that needs it most.

## Minor

- `pthread_atfork` handlers cannot be unregistered, and libwaste is loaded as a
  shared object through ctypes. A host that ever `dlclose`s it will jump into
  unmapped memory on the next `fork`. Worth a note next to `model_lock_init`.
- `flock` treats two open file descriptions in one process as independent, so a
  process can exclude itself in narrow cases where the `(dev, ino)` identity
  changes underneath it. The registry covers the normal path; this deserves a
  comment rather than code.
- `child_probe` execs `argv[0]`, which is fine from `run.sh` and fragile if the
  binary is ever invoked through `PATH`.

## What is right

I checked every return in `waste_open` after acquisition — there are exactly
three (plan, budget, model load) and all three release. `(dev, ino)` identity
instead of path strings, refcounting that preserves the documented
multiple-context behavior, close-on-exec with the `fcntl` fallback, and the
lock kept out of the token path are all the right calls. The comments explain
the failure that motivates them, which is the house style.

marcobambini:
Follow-up to the review: the RAM half of this now has its own home —
**#31**, "Two auto-budget opens on one machine each size against the whole
machine".

Tracing it back, the second-open case is *your* observation. #14's correction
— that the original 80.64 GiB paging sample came from a second WASTE model
open after budget selection — is where it enters the record, and #15
deliberately scoped itself to pressure already present at `waste_open`. So the
gap this PR is aimed at is real and was left open on purpose; it is the key
that I do not think holds, not the concern.

#31 sets out what is already settled (capacity via `min(physical, cgroup)`;
instantaneous `MemAvailable`/`memory.current` refuted in `src/memory.c`), the
question 0.6.5's changelog left open in as many words — whether pressure
should trim the working-set multiplier rather than the ceiling — and gates in
cheapest-first order, starting with actually measuring two concurrent K3 opens
on a 64 GB machine, which has never been done on purpose. There is a table
there showing why the inode is wrong on three of four configurations,
including the different-container case a lock cannot see.

If you want to take gate 1 or 2 on the GN100, that is the part of this whose
answer nobody currently has. Nothing in #31 blocks this PR: the errno fix and
the inverted default stand on their own.

mccoyspace:
Thanks for the detailed review. I agree with the distinction between optional
container ownership and RAM accounting, and have revised the single commit at
`bfd78c8` accordingly:

- inverted the API and flags to opt-in `exclusive_open` / `--exclusive-open`;
- made every non-contention locking failure proceed without ownership, with a
  search-only-directory regression case;
- bumped to 0.6.7 and added the ABI/behavior CHANGELOG entry;
- retained `test_lock` diagnostics and report synthetic-container failure as
  SKIP;
- added actionable `WASTE_E_BUSY` guidance in both CLI and server;
- added the `pthread_atfork`/`dlclose` and changing-identity comments; and
- changed the child probe to `execlp` for PATH invocation.

Current Mac validation is 31 passed, 0 failed, 13 skipped, including the
focused lock test, plus all 168 server checks. I also manually verified the
CLI/server contention advice and that a default open still succeeds while an
independent process holds the directory lock.

I have kept the RAM-budget question in #31 separate. We can add the GN100 gate
1/2 measurements there later without coupling them to this PR.

marcobambini:
Merged, as 0.6.8 rather than 0.6.7.

Every point from the review landed, and the second one landed better than I
asked for. Moving the over-wide-slice check into the same refusal mechanism
leaves one refusal point instead of a condition re-derived at the call site,
which is the version I should have proposed. Reporting a synthetic-container
failure as SKIP rather than as a pass is the `run.sh` discipline applied
correctly and without being asked.

## What I changed on the way in

0.6.7 shipped from `main` while this was in review, so the entry here described
a release that already exists. I retargeted the section to **0.6.8** and
rewrote it to cover what the release actually contains — the DeepSeek converter
(#26), the MLA rotary fix (#27) and this lock — with one ABI paragraph naming
every struct that moved rather than only `waste_cfg`. `WASTE_VERSION_PATCH`
went with it, per `CLAUDE.md`'s rule that the changelog moves in the same
commit as the version. Nothing of yours changed in substance; the tree is your
`bfd78c8` plus that resolution.

I also wrote the premise down in the changelog, because it was in the review
thread and nowhere in the repository: this is host policy, container identity
is a poor proxy for memory oversubscription, and the budget question is #31.
Anyone reading `--exclusive-open` in two years should find that next to the
flag rather than have to reconstruct it.

## Verification on the merge

Against real containers, not the synthetic path:

- `tests/run.sh` with `WASTE_REF_MODEL` on a default VQ3R Kimi-Linear:
  **54 passed, 0 failed, 2 skipped**, K3 checks included, with the ownership
  check among them.
- `test_lock` against a real 18 GB container: PASS.
- Serve suite: 211/211. The ctypes mirror carries both `exclusive_open` and
  0.6.7's `working_set_bytes`, so the two struct moves did not collide.

## The part I keep coming back to

The test. Verifying ownership with an independent `flock` rather than trusting
the library's own return, and using `exec` rather than relying on fork's copied
registry, is the right paranoia for a mechanism whose whole job is to be
believed by a second process. The `pthread_atfork` child handler closing
inherited descriptors before the copied registry can be consulted, with the
`dlclose` hazard written down beside it, is the kind of thing that is invisible
until it is wrong.

Separately: your GB10 results in #11 have moved that issue further than
anything here could. I am answering it there.

Issue #36, opened 2026-08-11
UCRT64 + first real Kimi-Linear container on Windows: whole-model oracle PASS on AVX2, and five gaps no CI job can reach


GTSUltear:
---

## Context

Built and ran WASTE **v0.6.6 (`df54209`) natively on Windows 11 x86_64** under MSYS2
**UCRT64**, then took Kimi-Linear-48B-A3B-Instruct all the way through: 91.5 GB of source
safetensors, a full `convert.py` run, and generation from the resulting 18 GB container.

This complements #3 rather than repeating it:

| | #3 (v0.6.0) | this run |
|---|---|---|
| subsystem | MinGW-w64 / Git-Bash; CI now MINGW64 | **UCRT64** |
| ISA | AVX-512 (Zen 4) | **AVX2 (Zen 2, Ryzen 5 3600X)** |
| oracle scope | KDA kernel (`kda_ref.py`) | **whole-model logits (`kimi_ref.py`)** |
| container | none (synthetic) | **real 18 GB, converted on this machine** |

That last row is why this is worth sending, and reading `ci.yml` on `main` sharpened the
point more than expected:

- `windows-native` uses **MINGW64** and installs no `uv`; it runs `bash tests/run.sh`
  with no argument, so the synthetic container.
- The only job that installs `uv` is the Linux `build` job, and it runs
  `tests/run.sh /nonexistent-container` — deliberately synthetic.
- **No job on any platform runs the suite against a real container.**
- `tools/diskbench.c` is never compiled in CI at all, on any platform.

So the checks that skip without weights — the budget check, the oracle diff, the
container round-trip — have never executed anywhere. **Every gap below lives in that
space**: each needs a real download, a real conversion, or a container to reach. None is
in `src/`; the engine came out clean.

Tested on **v0.6.6**. Every gap below re-verified against `main` at `2056d44` — i.e.
after v0.6.7 — and all five are still present. Note v0.6.7 reworked the automatic budget
resolver, so the cache measurements further down describe v0.6.6 behaviour.

Environment: AMD Ryzen 5 3600X (Zen 2, 6C/12T), 64 GB DDR4-3200, Sabrent Rocket NVMe4
1 TB, NTFS. gcc 16.2.0 (Rev3, MSYS2), GNU Make 4.4.1, CPython 3.12.8 (MSVC) in a venv
with torch 2.13.0.

---

## Results

[...]

PR #40, opened 2026-08-13
test: cover fp8 block-scale dequantization

### LINKED ISSUE #39: The fp8 block-scale mapping has no test, and the only stub naming it replaces it with the identity
`unblock_scale` in `tools/mxfp4.py:56` applies the per-tile scales an fp8
checkpoint ships alongside its weights. It landed with #26 and has **no test**.

The only place its name appears under `tests/` is a stub that replaces it with
the identity:

```python
# tests/test_convert_resume.py:103
mx.unblock_scale = lambda q, scale, block: q
```

That stub is correct for what it is there for — resume never reaches a
dequant, and `convert.py` imports the name, so it has to resolve or every
check in that file fails at module load. But it means the function's real
behaviour is unexercised, and a regression that made `unblock_scale` return
`q` unchanged would pass the entire suite.

## Why this one and not some other untested helper

Because of how it fails. Getting the block mapping wrong does not raise, does
not change any shape, and does not produce anything that looks wrong
downstream — it applies each scale to the wrong rows and yields plausible
weights. The docstring on the function makes the case itself:

> 300 rows with 3 scale rows admits both 128 (the truth, with a partial last
> tile) and 100 (a clean split), and the wrong one silently applies each scale
> to the wrong rows. The shapes agree in both readings, so nothing downstream
> notices.

A conversion is hours long and this runs at the front of it. The symptom would
surface as VQ reconstruction error nobody could attribute to a cause.

## What is asked

@fab2s — #26's description covers exactly this, and says it is already
written:

> `tools/../test_fp8_blocks.py` covers the block mapping on synthetic tensors
> (aligned tiles, partial tiles on both axes, a missing companion, a gross
> size mismatch) and cross-checks one real tensor bit-exactly against an
> index-array dequant; it also documents the case that cannot be caught, a
> compatible-but-wrong tile size, which is why the value is read from config.

It did not make it into the diff, and I asked for it in the merge comment.
Filing it here so it is tracked rather than sitting in a closed PR. Under
`tests/`, wired into `tests/run.sh`'s converter section next to
`test_convert_resume.py`, and it should **SKIP rather than fail** where torch
is unavailable — `run.sh`'s rule is that a missing prerequisite says SKIP,
loudly, and is never a silent pass.

The four synthetic cases plus the bit-exact cross-check are the right set. Two
things worth adding if they are not already in what you have:

- **A partial tile on the last row *and* column together.** `want` is computed
  with a ceiling on each axis independently and the result is cropped with
  `[:M, :N]` after two `repeat_interleave`s, so the corner is the place where
  an off-by-one in either axis would hide.
- **The two readers agreeing.** `ST.__init__` (`mxfp4.py:94`) and
  `ShardReader.__init__` (`convert.py:219`) each parse
  `quantization_config.weight_block_size` out of `config.json` separately, and
  each falls back to `(128, 128)`. Two copies of one rule drift; a check that
  both return the same tensor for the same input pins them together. #26's own
  description notes that `ShardReader` "is a second reader and was easy to
  miss", which is the same observation from the other side.

No urgency and no obligation — if you would rather not, say so and I will
write it from the description above; you have already done the part that
needed the checkpoint. What should not happen is it staying untested, given
that this is the one path in #26 that fails silently.

Related: #26.

tomatotomata:
Resolves #39

## What changed

- Added `tests/test_fp8_blocks.py` with aligned tiles, a partial final row and column, missing scale companions, and gross shape mismatch cases.
- Added a bit-exact index-based reference check for the block mapping.
- Asserted that `mxfp4.ST` and `convert.ShardReader` agree when both read the configured block size.
- Wired the test into `tests/run.sh`, with an explicit SKIP when torch is unavailable.

## Validation

- `python tests/test_fp8_blocks.py` passes all 5 checks.
- `python tests/test_convert_resume.py` passes.
- `python tests/test_convert_chat.py` passes.
- `git diff --check` passes.

The test stays synthetic and does not require model weights or a multi-hour conversion.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1342,10 +1342,19 @@ head_ "converter"
 # mode, so keep its tile mapping under a small synthetic test instead of
 # relying on a multi-hour model conversion. Exit 77 is an explicit skip when
 # torch is unavailable, never a pass.
-if ! command -v python3 >/dev/null 2>&1; then
-    sk "fp8 block-scale mapping" "python3 not installed"
+#
+# Through uv, like every other torch checker here: torch is not a dependency
+# of this repo and is not a system package on the machines that run this. As
+# bare `python3` the whole check reported "torch not installed" and skipped
+# everywhere, CI included — where the Linux job is the one that installs uv,
+# so this is exactly where it does get to run.
+if ! command -v uv >/dev/null 2>&1; then
+    # The guard is on uv rather than python3 for the same reason: without uv
+    # run_uv exits 127 and the catch-all below would call that a failure.
+    sk "fp8 block-scale mapping" "uv not installed"
 else
-    out=$(python3 tests/test_fp8_blocks.py 2>&1); rc=$?
+    out=$(run_uv run --quiet --with torch --no-project \
+              python tests/test_fp8_blocks.py 2>&1); rc=$?
     case "$rc" in
         0)  ok "fp8 block scales, partial tiles, missing companions, and reader agreement" ;;
         77) sk "fp8 block-scale mapping" "torch not installed" ;;
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1202,6 +1202,21 @@ fi
 # ------------------------------------------------------------ converter ----
 head_ "converter"
 
+# The fp8 converter path has two readers and a silent shape-preserving failure
+# mode, so keep its tile mapping under a small synthetic test instead of
+# relying on a multi-hour model conversion. Exit 77 is an explicit skip when
+# torch is unavailable, never a pass.
+if ! command -v python3 >/dev/null 2>&1; then
+    sk "fp8 block-scale mapping" "python3 not installed"
+else
+    out=$(python3 tests/test_fp8_blocks.py 2>&1); rc=$?
+    case "$rc" in
+        0)  ok "fp8 block scales, partial tiles, missing companions, and reader agreement" ;;
+        77) sk "fp8 block-scale mapping" "torch not installed" ;;
+        *)  no "fp8 block-scale mapping"; printf '%s\n' "$out" | grep -E "FAIL|Error|Traceback" | head -5 ;;
+    esac
+fi
+
 # Resume is the one converter behaviour that cannot be checked by looking at
 # a finished container: it is about the partial states a crash leaves. The
 # quantizer is stubbed out, so this needs neither torch nor source weights.
--- a/tests/test_fp8_blocks.py
+++ b/tests/test_fp8_blocks.py
@@ -6,6 +6,16 @@
 These tests use small synthetic tensors so they do not need model weights. A
 missing torch installation is an explicit skip, matching tests/run.sh's rule
 that unavailable prerequisites must never look like a pass.
+
+The one case nothing here can catch, stated because it is the reason the tile
+size is read from the checkpoint's config rather than inferred from the two
+shapes: a *compatible but wrong* block size. 300 rows against 3 scale rows
+admits both 128 (the truth, with a partial last tile) and 100 (a clean split).
+Both satisfy the shape check below, both produce a tensor of the right size,
+and the wrong one applies every scale to the wrong rows. No assertion over
+shapes can separate them, which is why `unblock_scale` takes `block` as an
+argument instead of deriving it — the check that matters happened before this
+file was reached.
 """
 
 import json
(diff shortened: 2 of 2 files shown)


marcobambini:
Reviewed. The test is good and I checked it by trying to break it rather than
by reading it — all four mutants I planted in `unblock_scale` were caught:

| mutation | result |
|---|---|
| return `q` unchanged (the exact regression the resume stub simulates) | **killed** |
| transpose the block dims, `repeat_interleave(bn, 0)…(bm, 1)` | **killed** |
| drop the `[:M, :N]` crop | **killed** |
| remove the scale-shape guard | **killed** |

`expected_dequant` being an independent index-array implementation rather than
a second call to the thing under test is what gives it that power, and the 3x5
against block (2,3) case puts a partial tile on both axes at once, which is the
corner where the two independent ceilings and the crop meet. Both were what #39
asked for, and the reader-agreement check pins `ST` and `ShardReader` together,
which is the drift #26 itself warned about.

Two things before this comes out of draft.

## 1. It never runs, on any machine that tests this project

`run.sh` invokes it as bare `python3`, and torch is not a system package here —
per `CLAUDE.md` it is never a repo dependency and every torch checker goes
through `uv run --with torch --no-project`. So on my machine, with uv installed
and every other torch check running, this reports:

```
converter
  SKIP  fp8 block-scale mapping — torch not installed
```

It skips in CI too, and that is the part that decides it: the Linux `build` job
is the one job that installs uv, so with the right wiring this check would
actually execute there — and #39 exists because the fp8 path is the one with a
silent failure mode. A test that never runs is the same amount of coverage as
no test.

Confirmed the fix, with the `command -v uv` guard the other torch sites use:

```bash
if ! command -v uv >/dev/null 2>&1; then
    sk "fp8 block-scale mapping" "uv not installed"
else
    out=$(run_uv run --quiet --with torch --no-project \
              python tests/test_fp8_blocks.py 2>&1); rc=$?
```

`SKIP` becomes `PASS` with that change and all five checks execute. `run_uv` is
new on `main` since you opened this — it puts a deadline on uv, which returns
124 on a hang, and your `*` case already treats that as a failure. Note the
guard has to be on `uv` rather than on `python3`: without uv present, `run_uv`
exits 127 and your `*` branch would report FAIL where it should SKIP.

## 2. One reader's guard is uncovered

`test_missing_scale_companion_is_rejected` exercises `ST` only. I deleted
`ShardReader`'s own `if sname not in self.wm: raise KeyError` — the one in
`convert.py` — and made it return the tensor unscaled instead. **The suite
stayed green.** That is precisely the silent-wrong-answer shape this file
exists to prevent, in the reader #26's description called out as "a second
reader and was easy to miss".

`write_safetensors_model(root, block, include_scale=False)` already builds the
directory; it needs the `ShardReader(root).get("weight")` half added alongside
the `ST` one.

## Smaller

- Rebase on `main`: it has moved a fair bit today (0.6.8, plus five fixes from
  #36 that touch `run.sh`). It still merges clean, but `run_uv` only exists
  after that.
- The `bytes(...view(torch.uint8).flatten().tolist())` round-trip works but
  `.numpy().tobytes()` is not available without numpy, so if you want it
  shorter, `bytes(q.view(torch.uint8).flatten().tolist())` is already the
  portable form — no change needed, just noting I checked it rather than
  assuming.
- #26's description mentioned the test would also *document* the case that
  cannot be caught — a compatible-but-wrong tile size, which is why the value
  is read from config rather than inferred. A comment saying so would put the
  reasoning next to the checks that cannot express it.

Nothing here is structural, and the mutation results say the core is right.
Happy to take it as soon as it runs. Thanks for picking this up — it was filed
this morning against a different contributor's PR and you got there first.

marcobambini:
Both points fixed and pushed to this branch —
[494f53e](https://github.com/sqliteai/warp/commit/494f53e), on top of your
commit, which is unchanged. Rebased on `main` at the same time, since it moved
a good deal today.

**1. It now runs.** Through `run_uv run --with torch --no-project`, with the
guard on `uv` rather than `python3`. The suite goes from 56 passed / 0 failed /
2 skipped to **57 / 0 / 2** against real containers, and — the part that
actually decides it — here is the Linux CI job, which is the one that installs
uv:

```
converter
  PASS  fp8 block scales, partial tiles, missing companions, and reader agreement
```

That line did not exist before. Under bare `python3` it read
`SKIP — torch not installed`, everywhere, CI included.

**2. Both readers.** `test_missing_scale_companion_is_rejected` now loops over
`ST` and `ShardReader`. The mutation that survived in my review — deleting
`convert.py`'s own `raise` so it returns the tensor unscaled — is killed.

One thing I checked and deliberately did **not** change: the same mutation
against `ST` still survives, and it should. Without that guard, `raw()` raises
`KeyError: 'weight_scale_inv'` anyway, so the contract the test asserts —
refuses, and names the missing tensor — holds either way. That guard buys a
better *message*, not a different decision, and pinning its exact prose would
test the wording rather than the behaviour.

I also wrote down the case nothing here can catch, since #26 described it and
it is the reason `unblock_scale` takes `block` as an argument instead of
deriving it: 300 rows against 3 scale rows admits both 128 and 100, both pass
every shape check, and the wrong one applies each scale to the wrong rows.

CI is 9/9 on the branch. I approved the workflow run — it needed a maintainer
for a first-time contributor, and I had read and executed the whole diff
locally by then.

**Over to you to take it out of draft**; I did not want to flip that on your
PR. Once you do it can go straight in.

For the record, your checks were the right ones and they were right on the
first try: `expected_dequant` as an independent index-array implementation
rather than a second call to the code under test is what gave the file its
teeth, and 3x5 against block (2,3) puts a partial tile on both axes at once,
which is the corner where the two ceilings and the crop meet. Everything above
is wiring and one missing reader. Thanks for picking this up the morning it was
filed.

marcobambini:
Merged, and #39 closed with it.

CI 9/9 on the merge commit, and the check runs rather than skips on the Linux job — which was the whole difference between this being coverage and being a file.

Thanks. Filed in the morning, correct checks by the afternoon, and the mutation results were the part that made it easy to take: four planted bugs in `unblock_scale` all caught, including the one the resume stub had been silently simulating since #26 landed.

Issue #37, opened 2026-08-13
Gate 7: Strix Halo (128 GB, Zen 5 AVX-512/VBMI) dataset for the fine cache sweep


Lrrr908:
We have a 128-GB Ryzen AI Max+ 395 Strix Halo Linux system with Zen 5 AVX-512/VBMI. We can run the open Gate-7 fine cache sweep at 200+ tokens and characterize the larger-memory side of the paging cliff.

Is Gate 7 already being worked, and would a Strix Halo dataset be useful? If yes, what exact command/configuration/output format would you prefer?

For reference, the gate as we understand it: docs/GATES.md @ 2056d44f (lines 318–375) — tests/sweep.c with cache=, a fine grid between 287 and 1498 slots, over a long generation (200 tokens or more), one process, interleaved arms per LEARNED.md §38. Kill criterion: if at 200+ tokens a 3–4 GB cache is no longer within ~10% of floor + 1x, the rule stands as written.

marcobambini:
Gate 7 is open, **nobody is working it**, and yes — a dataset would be useful.
Answering the command question below, but two things come first, because one
of them is worth more than Gate 7 and costs minutes instead of an hour.

## Your machine can do something no machine has ever done here

`docs/BACKENDS.md` records AVX-512 as **"compiled and dispatched, never
executed"**. That is literal: `src/simd_avx512.c` has been compiled and
CPUID-dispatched since 2026-07-28 and no machine this project can reach has
ever run an instruction from it. The container CPU reports `AVX512F=0`, and
so does `qemu-x86_64 -cpu max` under Rosetta.

So before any sweep, please just run the suite:

```bash
make && make test && tests/run.sh /path/to/container
```

The line to report is ***SIMD backend matches the CPU baseline***. It runs
`WASTE_BACKEND=cpu` against the dispatched path and compares logits, so on a
Zen 5 it is the first execution of those kernels anywhere. It costs a couple
of minutes and it either confirms four months of untested code or finds a
bug in it. `./waste info CONTAINER` should print `backend AVX-512`; if it
prints `AVX2` or `CPU`, that is itself the result and worth reporting.

Kimi-Linear (19 GB) is fine for this — it does not need K3.

## And your machine has VBMI, which #38 needs

I opened #38 an hour ago: VQ4P's apply is vectorized behind an `#ifdef` in
`model.c` and never moved behind the dispatch table, so x86 has no kernel for
[...]

PR #41, opened 2026-08-20
Add vq_rows_p6 dispatch slot and AVX-512 VBMI kernel

### LINKED ISSUE #38: VQ4P's apply is an #ifdef in model.c, not a dispatch slot — which is why x86 has no kernel
`vq_rows_p6` — the VQ4P apply — is vectorized behind an `#ifdef` in
`src/model.c:2158` rather than behind the dispatch table, and that is the
structural reason x86 has no kernel for it rather than a narrow one.

```c
#if (defined(__ARM_NEON) || defined(__aarch64__)) && !defined(WASTE_P6_SCALAR)
            if (nr == VQ_TILE) {
                /* four vqtbl4q_s8 per 16 rows over a vld3q_u8 deinterleave */
            } else
#endif
            {
                /* one row at a time, four scalar table loads per row */
            }
```

The two range kernels that carry the rest of the arithmetic — `mvq_rows_f32`
and `lutb_range` — moved behind `waste_kernels` on 2026-07-28 precisely so
that one binary adapts from CPUID and the ISA lives in its own translation
unit. This one did not move with them, so an x86 build takes the portable
`else` branch: `T[j0] + T[en + j1] + T[2*en + j2] + T[3*en + j3]`, per row.
That is not a narrower kernel, it is the absence of one, on the same build
where VQ3R's table gets `simd_avx512.c`.

Successor to #32, which is closed: its documentation half landed in
[48b8afa](https://github.com/sqliteai/warp/commit/48b8afa), and this is the
half that needs code.

## Scope: who actually meets this

Narrower than #32 claimed, and worth stating before anyone prioritises it.
#32 said VQ4P is what `convert.py` produces by default. It is not —
`--index-bits` defaults to 8 and `--stages` to 3, so a default conversion is
VQ3R, and `convert.py:903` refuses `--index-bits 6` without
`--stages 4 --entries 64`. **So this is not the path on every x86 build.**

What makes it worth doing anyway is *who* the affected population is. That
shape is exactly the 64-entry crossover `docs/LEARNED.md` §50 lists as not
measured. So the people most likely to run VQ4P on x86 are the people trying
to answer that open question — and what they would measure is a scalar apply
against a vectorized one, then report an ISA difference as a table-size
result. A wrong number that looks like an answer costs more than a slow path
that announces itself.

## Why a kernel is possible here and is not for VQ3R

§41 has the arithmetic. VQ3R's 256-entry stage table is 256 bytes — sixteen
vector registers of the thirty-two that exist — so it fits no byte-permute
primitive on any ISA, and blocking does not make it fit. VQ4P spends the same
24 bits per 8-weight vector at 4x64, and a 64-entry stage table is **64 bytes,
four registers**, which is exactly what `vqtbl4q_s8` addresses. The premise
that made the gather scalar is false for this format.

The x86 analogue is **AVX-512 VBMI's `vpermi2b`**: 128 bytes indexed across
two zmm registers, so 64 entries fit with room to spare. Zen 4 and later, Ice
Lake and later.

## The work, in the order that makes it reviewable

1. **Add the dispatch slot first, with the ARM body moved into it unchanged.**
   `waste_kernels` has no entry for this kernel; add one taking
   `(int b, int e, void *arg)` like the other two, move the NEON body out of
   `model.c` into its own translation unit, and register it from the ARM path.
   This step should be **bit-identical on ARM by construction** and is worth
   landing on its own — it is reviewable without an x86 machine, and it is
   where a mistake would be cheapest to find.

   It also improves the self-check. `-DWASTE_P6_SCALAR` is a compile-time
   lever today; once the kernel is a slot, `WASTE_BACKEND=cpu` reaches the
   portable path at runtime, which is the lever the existing suite check
   already uses.

2. **`WASTE_CPU_AVX512VBMI` does not exist yet.** `src/waste_backend.h:56-57`
   has `AVX512F` (bit 11) and `AVX512BW` (bit 12); VBMI is CPUID leaf 7
   subleaf 0, ECX bit 1, and `src/backend.c:123` is where its siblings are
   detected. Gate on it specifically — `AVX512F && AVX512BW` at
   `backend.c:190` is not sufficient for `vpermi2b`.

3. **Then the kernel**, in `src/simd_avx512.c`, which already exists.

## Acceptance

- **Bit-identity against the portable path, not "close".** §43 says why the
  bar is that high: an int8 table makes the engine discontinuous, so an
  approximate match is not a match. It is reachable — accumulation is int16
  within a `WASTE_VQ_LUT_BLK` block and only folds to fp32 at the end, so
  there is no float reordering for a different lane arrangement to expose.
  The ARM path already claims this against `-DWASTE_P6_SCALAR`; x86 has to
  clear the same bar against the same reference.
- **The check that decides it is `tests/run.sh`'s *SIMD backend matches the
  CPU baseline***, which runs `WASTE_BACKEND=cpu` against the dispatched path
  and compares logits.
- **It has to run on an `--index-bits 6` container.** The suite's default
  `WASTE_REF_MODEL` is a VQ3R conversion, which does not execute this kernel
  at all; per `CLAUDE.md`, running the suite on the wrong container shape is
  exactly how a load path once stayed broken through green runs. Build one
  with `convert.py --index-bits 6 --stages 4 --entries 64`.

## Unmeasured in both directions

Stated before anyone spends a weekend on it:

- The path may be bound by the **deinterleave** (`vld3q_u8` on ARM, three
  interleaved index bytes per row) rather than by the lookup, in which case
  the table primitive buys much less than it looks like it should.
- **Cross-lane byte permutes are not uniformly cheap** across x86
  microarchitectures, and `vpermi2b`'s cost differs between Ice Lake and
  Zen 4/5.

This is a hypothesis with an obvious first experiment, not a plan.

## Nobody here can execute it

This repository has no x86 machine that is not emulated, which is also why
`docs/BACKENDS.md` records AVX-512 as "compiled and dispatched, never
executed", and why the gap was found by reading rather than by measurement.
**Whoever writes this is also the only person who can test it**, so the
measurement matters more than the patch — and step 1 above is deliberately
carved out as the part that can be reviewed here.

Related: #11 (a GPU VQ4P implementation exists off-repo and found that
building the table dominated applying it — a caution about where the time
actually goes, on a different vehicle), and #36 (the suite's non-synthetic
path does not run in CI on any platform, which is the reason a container-shape
requirement in the acceptance criteria above cannot be enforced by CI today).

Ultron09:
Move VQ4P apply out of an \#ifdef\ in \src/model.c\ into the engine's kernel dispatch table (\waste_k.vq_rows_p6\), matching the pattern of other arithmetic kernels in sqlite-vector/WARP.

### Summary of Changes
- **Dispatch slot**: Added \oid (*vq_rows_p6)(int b, int e, void *arg);\ to \waste_kernels\ in \src/waste_backend.h\.
- **CPU baseline**: Exported portable C baseline \waste_vq_rows_p6\ in \src/model.c\ and registered it in \waste_kda_register_cpu\ in \src/kda.c\.
- **NEON backend**: Moved NEON vectorized implementation \q_rows_p6_neon\ to \src/kda_neon.c\ and registered it in \waste_kda_register_neon\.
- **CPUID feature detection**: Added \WASTE_CPU_AVX512VBMI\ capability bit in \src/waste_backend.h\ and detection for CPUID leaf 7 subleaf 0 ECX bit 1 in \src/backend.c\.
- **AVX-512 VBMI kernel**: Implemented \q_rows_p6_avx512\ in \src/simd_avx512.c\ using \_mm512_permutexvar_epi8\ for 16-row vector unpack and 64-byte stage table lookups, registered in \waste_register_avx512\.
- **Build flags**: Added \-mavx512vbmi\ to CFLAGS for \src/simd_avx512.o\ and \src/simd_avx512.pic.o\ in \Makefile\.

### Verification
- Built full suite via \make && make test\.
- Ran \	ests/run.sh\: all 42 checks passed (0 failed).
- Verified bit-identity between \waste_vq_rows_p6\ (CPU baseline) and the dispatched kernel across randomized fixtures.

Closes #38

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,7 +186,7 @@ src/metal.o: src/metal.m
 # so helpful as warning that the flag went missing. Invisible on ARM,
 # where these translation units are not in SRC at all.
 src/simd_avx2.o:   override CFLAGS += -mavx2 -mfma
-src/simd_avx512.o: override CFLAGS += -mavx512f -mavx512bw
+src/simd_avx512.o: override CFLAGS += -mavx512f -mavx512bw -mavx512vbmi
 
 all: waste$(EXE) libwaste.a libwaste.$(SOEXT) libwastevq.$(SOEXT)
 
@@ -216,7 +216,7 @@ src/metal.pic.o: src/metal.m
 	$(CC) $(CFLAGS) $(PICFLAG) -fobjc-arc -c -o $@ $<
 
 src/simd_avx2.pic.o:   override CFLAGS += -mavx2 -mfma
-src/simd_avx512.pic.o: override CFLAGS += -mavx512f -mavx512bw
+src/simd_avx512.pic.o: override CFLAGS += -mavx512f -mavx512bw -mavx512vbmi
 
 libwaste.$(SOEXT): $(SHOBJ)
 	$(CC) $(CFLAGS) -shared -o $@ $^ $(SHLDFLAGS) $(LDLIBS)
--- a/src/backend.c
+++ b/src/backend.c
@@ -122,6 +122,7 @@ uint32_t waste_cpu_features(void)
             if (ymm_ok && (r[1] & (1u << 5)))  f |= WASTE_CPU_AVX2;
             if (zmm_ok && (r[1] & (1u << 16))) f |= WASTE_CPU_AVX512F;
             if (zmm_ok && (r[1] & (1u << 30))) f |= WASTE_CPU_AVX512BW;
+            if (zmm_ok && (r[2] & (1u << 1)))  f |= WASTE_CPU_AVX512VBMI;
             run_cpuid(7, 1, r);
             if (ymm_ok && (r[0] & (1u << 4)))  f |= WASTE_CPU_AVX_VNNI;
         }
--- a/src/kda.c
+++ b/src/kda.c
@@ -114,14 +114,16 @@ void waste_rmsnorm_gated(int C, const float *x, const float *gate,
  * overwrite only what they implement (sqlite-vector discipline). */
 void waste_mvq_rows_f32(int b, int e, void *p);
 void waste_lutb_range(int lo, int hi, void *p);
+void waste_vq_rows_p6(int b, int e, void *p);
 
 const char *waste_kda_register_cpu(waste_kernels *t)
 {
     t->kda_step = waste_kda_step;
     /* The portable range kernels live in model.c, next to the code that
-     * builds their arguments; an ISA backend overwrites these two. */
+     * builds their arguments; an ISA backend overwrites these. */
     t->mvq_rows_f32 = waste_mvq_rows_f32;
     t->lutb_range = waste_lutb_range;
+    t->vq_rows_p6 = waste_vq_rows_p6;
     t->short_conv_step = waste_short_conv_step;
     t->rmsnorm_gated = waste_rmsnorm_gated;
     return "CPU";
--- a/src/kda_neon.c
+++ b/src/kda_neon.c
@@ -13,6 +13,7 @@
 
 #include "kda.h"
 #include "waste_backend.h"
+#include "simd.h"
 
 #include <arm_neon.h>
 #include <math.h>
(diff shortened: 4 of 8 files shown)


Ultron09:
cc @marcobambini — PR is ready for review. Implemented both the VQ4P dispatch slot refactor and the AVX-512 VBMI kernel with all tests passing and verified for bit-identity against the baseline.

mfethe1:
Not a maintainer — but I have a native x86 box (Ryzen 7 3700X, Zen 2, Windows 11, MSYS2 UCRT64, gcc 16.2.0) stood up for #37/#38 work, so I built and tested this. Offering what the hardware can actually settle, and being explicit about what it can't.

## Result

Builds clean, no new warnings. Full suite: **44 passed, 0 failed, 13 skipped — byte-for-byte the same board as `main` on this machine.** No regression from the refactor.

**Caveat that matters more than the result:** on Zen 2 there is no AVX-512 and no VBMI, so I cannot execute `vq_rows_p6_avx512`. Everything below about that function is inspection, not measurement.

## The dispatch refactor looks right

- `waste_kda_register_cpu` is called unconditionally at `src/backend.c:175` before any ISA backend, and each backend overwrites only what it implements, so `waste_k.vq_rows_p6` can never be NULL. That was the failure mode I went looking for and it isn't there.
- Hoisting `vqp_arg` and the `P6_J*` macros into `simd.h` is the right call. The macros' own comment says they exist so the paths "cannot drift apart"; with three implementations that is now load-bearing rather than aspirational.

## The AVX-512 kernel, by inspection

I worked through the index math and it holds up:

`_mm512_maskz_loadu_epi8(0x0000FFFFFFFFFFFF, …)` zeroes bytes 48–63, so `unpack_mask`'s byte-63 slot contributes 0 and each dword becomes a clean 24-bit `b0 | b1<<8 | b2<<16`. From there `j0..j3` reduce exactly to `P6_J0..P6_J3`. Good.

Two things worth a comment in the source, because both look wrong on first read and aren't:

1. **`_mm512_permutexvar_epi8(j0, T0)` returns garbage in three of every four bytes.** `j0`'s dword lanes have zero upper bytes, so those select `T0[0]`. The following `slli_epi32(…,24)` / `srai_epi32(…,24)` discards them and sign-extends byte 0, which is what makes it correct. A reader who doesn't spot the shift pair will file a bug.

2. **The accumulator width differs from the other two paths** — `int32` here, `int16` in NEON and scalar. It is still bit-identical: two's-complement addition mod 2³² reduced mod 2¹⁶ at the end equals accumulating mod 2¹⁶ throughout. (And moot in practice, since the existing comment bounds the sum at 16256.) Given this project's standard is bit-identity rather than closeness, that equivalence is worth stating where someone will find it.

## The thing I'd want settled before merge

**This kernel is never executed by `make check`.**

`tools/make_test_container.py` has no `--index-bits` option and `tests/run.sh` never passes one, so the synthetic container is always `index_bits=8` and `vq_rows_p6` is unreachable in the suite. My 44/0/13 above — and any CI green on this PR — says nothing whatsoever about the code being added.

CLAUDE.md names this exact trap:

> the same applies to `--index-bits 6`, which exercises a different kernel and a different record fmt

> running the suite on one is how the Q4G load path stayed broken through green runs

I tried to close it myself and couldn't, which I think is the useful data point. `convert.py` needs `--index-bits 6 --stages 4 --entries 64` together (good error message, incidentally). I converted a Kimi-Linear layer that way, but `--layers 1` writes a manifest with one entry in `layers` against a `config` declaring 27, and the engine correctly refuses it as `malformed container`. So exercising this path at all requires a **full** conversion at `--index-bits 6` — hours, and source weights on disk. That is a high bar for a reviewer, and it means this kernel would land untested by anything automated.

Cheapest fix I can see: teach `make_test_container.py` `--index-bits 6` and add a VQ4P arm to the engine block in `run.sh`. Then the existing "SIMD backend matches the CPU baseline" check covers this for free on every platform. I'm happy to send that as a separate PR if it would help — it's orthogonal to your change and would give it a home to be verified in.

## One hardening suggestion, tested rather than asserted

`-mavx512vbmi` is added to `src/simd_avx512.o` for the whole translation unit, and that TU also holds `mvq_rows_f32_avx512` and `lutb_range_avx512` — which are dispatched on AVX512F+BW alone. Skylake-SP and Cascade Lake have F+BW and **no VBMI**, so in principle the compiler could place a VBMI instruction in a function that runs there.

I checked instead of speculating. Compiled your `simd_avx512.c` with your exact flags and disassembled:

```
$ gcc -O2 -Wall -Wextra -mavx512f -mavx512bw -mavx512vbmi -c src/simd_avx512.c
$ objdump -d simd_avx512.o | grep -E 'vpermb|vpermi2b|vpermt2b'
      5  <vq_rows_p6_avx512>:
```

Five VBMI instructions, all inside the guarded kernel, none anywhere else. **So this does not misbehave today** with gcc 16.2.0 at `-O2`.

But that outcome is the compiler's choice, not the source's — a different gcc, `-O3`, or a later edit to the neighbouring kernels could change it, and the failure would be a SIGILL on exactly the used-Xeon-Scalable hardware this engine is most attractive on. `__attribute__((target("avx512vbmi")))` on the single function, or moving it to its own TU, makes the guarantee structural. Your call whether that's worth it now or a TODO.

## Small one

`#if defined(__AVX512VBMI__)` means a build that doesn't get the Makefile's flag silently omits the kernel and falls back to scalar rather than failing. Fine as long as the Makefile is the only build path, but it's a quiet downgrade if anyone ever compiles this TU directly.

---

Happy to re-run anything on this box, or to send the `make_test_container.py --index-bits 6` change if you'd like this to have a test to sit behind. Nice work on the refactor — the dispatch-slot half is clearly right and useful independently of the VBMI kernel, which is what #38 said too.

Ultron09:
@mfethe1 Thank you for the outstanding review, hardware verification on Ryzen 7 / Zen 2, and disassembly analysis!

1. **Target Attribute & In-Source Documentation Added**:
   - Pushed commit `c3e9ae3` adding `__attribute__((target("avx512vbmi")))` directly to `vq_rows_p6_avx512` so the compiler's instruction generation for VBMI is structurally isolated to this kernel and cannot leak into neighboring F+BW functions in the translation unit on Cascade Lake / Skylake-SP.
   - Added clear explanatory comments in `simd_avx512.c` detailing why the `_mm512_permutexvar_epi8` + `slli_epi32`/`srai_epi32` shift pair cleanly discards unselected upper bytes and why 32-bit addition truncated mod $2^{16}$ is bit-identical to 16-bit accumulation.

2. **Test Coverage & Container Support**:
   - That point about `make_test_container.py` and `tests/run.sh` lacking `--index-bits 6` is spot on. If you'd like to open a PR adding `--index-bits 6` to `make_test_container.py` and wiring up the VQ4P engine check in `run.sh`, that would be hugely appreciated and give automated CI coverage for all three backends (AVX-512, NEON, and scalar baseline)!

Thanks again for the thorough review and testing!

mfethe1:
Opened the follow-up: #48 — `--index-bits 6` in `make_test_container.py`, a VQ4P engine section in `run.sh`, and `test_container` reading the `WQ_VQ4P` fmt byte. Default containers stay byte-identical (verified by sha256, so the rotary fixture's `container_sha256` gate is untouched); the new arm ran clean at exactly baseline+4 on both macOS arm64 (NEON) and Windows/Zen 2 (scalar), no prior PASS moved. Once this kernel's AVX-512 path lands, it gets the same CI coverage on Cascade Lake / Skylake-SP for free.

marcobambini:
Merged — and this repository has a VQ4P container, so the half that could be
verified here was verified rather than reasoned about.

**The dispatch-slot refactor is bit-identical on ARM against a real
`index_bits: 6` container.** Same 8 ids, same forward pass, merge-base against
this branch:

```
kimi-linear-vq4p (VQ4P, 4 stages / 64 entries, index_bits 6)   identical
kimi-linear-4x64 (CTRL, 4 stages / 64 entries, index_bits 8)   identical
```

That is the claim the PR makes about step 1 — the NEON body moved into a slot,
unchanged — and it now has a container behind it instead of an argument. The
VBMI kernel is a different matter: no machine here can execute an AVX-512
instruction, so it lands under the same standing caveat `docs/BACKENDS.md`
records for the rest of that file.

@mfethe1's review is why this went in as it stands. The `objdump` check instead
of a speculation about instruction placement, and the `__attribute__((target))`
that followed it, turn "gcc 16.2.0 at -O2 happens to do the right thing" into a
structural guarantee — and the failure it prevents is a SIGILL on Skylake-SP
and Cascade Lake, which is exactly the used-Xeon hardware this engine is most
attractive on.

## The test gap is real, and it is worse than "no coverage"

You are both right that `tools/make_test_container.py` has no `--index-bits`,
so the synthetic container is always 8-bit, `vq_rows_p6` is unreachable from
the suite, and a green CI run on this PR says nothing about the code it adds.
That should be closed and neither of you should be blocked on the other: send
the `make_test_container.py --index-bits 6` change whenever suits, it is
orthogonal and welcome.

But adding a VQ4P arm to *"SIMD backend matches the CPU baseline"* will not
work, and I can now say that from this side rather than quoting #36 at you.
Same two containers, dispatched path against `WASTE_BACKEND=cpu`:

| container | index bits | NEON vs CPU |
|---|---|---|
| CTRL | 8 | max diff `4.53e-06`, **0**/163840 over 1e-3 |
| VQ4P | **6** | max diff `0.414783`, **162233**/163840 |

@GTSUltear measured `1.22853` and 163297/163840 on AVX2 with the same
single-variable control. **Two different ISAs, same structure, same
conclusion** — so the divergence is not an x86 kernel defect, it is the int8
LUT quantization being discontinuous, and that check cannot pass on an
`index_bits: 6` container on any platform.

Which means the coverage this kernel needs is **bit-identity of the int8 table,
or of the routes, against the scalar path** — not a tolerance on logits, since
a genuinely broken kernel and a single entry rounding the other way both land
around 1. `LEARNED.md` §43 arriving somewhere concrete, as #37 put it. Worth
settling before `vpermi2b` gets a second kernel beside it.

Thanks to you both — one for the refactor and the target attribute, one for the
review that found the thing the CI could not.

Issue #32, opened 2026-08-08
The VQ4P apply is NEON-only, so §47's tuning does not travel to x86


marcobambini:
`src/model.c:2137` gates the vectorized VQ4P apply:

```c
#if (defined(__ARM_NEON) || defined(__aarch64__)) && !defined(WASTE_P6_SCALAR)
```

The vectorized side is four `vqtbl4q_s8` per 16-row tile — a 64-byte table
lookup — over a `vld3q_u8` deinterleave, with widening adds into int16. An x86
build takes the `else` branch: one row at a time, four scalar table loads per
row. It is not a narrower kernel, it is the absence of one — and on the same
build VQ3R's table gets `simd_avx512.c`.

Reported by @fab2s in #11, while trying to run the 64-entry crossover
`docs/LEARNED.md` §50 left open. Confirmed here by reading rather than by
measurement: this repo has no x86 machine that is not emulated, which is also
why `docs/BACKENDS.md` records AVX-512 as "compiled and dispatched, never
executed".

## What is being assumed because of it

- **The 64-entry crossover cannot be measured on an x86 CPU at all.** §50 left
  open whether VQ4P's 72 KB int8 table inverts the LUT-versus-reconstruct
  result it measured at 256 entries. On x86 that comparison puts a scalar apply
  against a vectorized one, so the number would be answering an ISA question
  and not a table-size one.
- **§47's tuning table is an ARM result.** `WASTE_XPAR`, `WASTE_XPAR_BATCH`,
  `WASTE_P6_CHUNK` and the thread counts were measured against the vectorized
  apply, and `CLAUDE.md` carries them as engine-wide guidance. On x86 the kernel
  underneath them is a different kernel, and the inversion between Kimi-Linear
  and K3 has never been retested there.
- **VQ4P is the format `convert.py` produces by default**, so this is the path
  on every x86 build, not a corner case.

## What might close it

`vqtbl4q_s8` looks up 64 bytes across four registers. The AVX-512 analogue is
**VBMI's `vpermi2b`**, which indexes 128 bytes across two zmm registers — the
same primitive with room to spare for a 64-entry table. Zen 4 and later, and
Ice Lake and later, have VBMI. `src/simd_avx512.c` already exists as the place
it would go, and dispatch resolves once at init, so this is a slot to fill
rather than an `#ifdef` in `model.c`.

Unmeasured in both directions, and worth stating before anyone spends a
weekend on it: the path may still be bound by the deinterleave rather than by
[...]

PR #42, opened 2026-08-21
A missing cmp or python3 said the engine was broken, not the tool absent

mfethe1:
`tests/run.sh:10` states the contract:

> never treats a missing prerequisite as a pass — it says SKIP, loudly.

and the `curl` guard says it again, in more words, for exactly this reason:

> The worker the downloader generates is curl, so without curl these three report the downloader broken when what is missing is a tool — **which is the one thing this suite says it must never do.**

`cmp` and the `python3` behind `range_server.py` were not held to that. This brings them into line.

## Evidence

Captured on Windows 11 / Ryzen 7 3700X (Zen 2) / MSYS2 UCRT64 / gcc 16.2.0. The before side is the merge-base `7f1fbba` in a sibling worktree, the after side is this branch; each tool was hidden by moving its binary aside for the duration of *both* runs, so the two sides differ only in the commit. `LC_ALL=C`, `TERM` and `COLUMNS` pinned, identical `grep` filter applied to both.

**`cmp` (diffutils) not installed** — before, eight checks report FAIL and name the engine. `expert cache changes results` is the report of a bit-identity violation in the expert cache, and it fired because `/usr/bin/cmp` was absent. After, the same eight report SKIP and name the missing tool. The pass count is unchanged at 36; only the eight false failures move.

| Before | After |
| --- | --- |
| 8 FAIL lines, 36 passed 8 failed 13 skipped | 8 SKIP lines naming cmp, 36 passed 0 failed 21 skipped |

**`python3` not installed** — before, four checks report `range server did not start`, blaming the download script for a missing interpreter. Note that `convert.py` and `serve`, which *are* guarded, correctly SKIP in the very same frame: the inconsistency is visible without leaving the image. After, all five download-script checks SKIP and name the tool, consistent with the checks beside them.

| Before | After |
| --- | --- |
| 4 FAIL range server lines beside correctly-skipped convert.py checks | all 5 download checks SKIP naming python3 |

## Why this matters more than a cosmetic label

Every one of those FAIL lines names the engine or the download script. `expert cache changes results` is exactly what this suite is *for* — and in a log it is indistinguishable from the real defect. `mla_use_nope: false was read as NoPE and skipped the rotation` names the precise bug #27 fixed.

A first-time Windows contributor's first run of this suite currently accuses the expert cache of returning different logits, on a clean checkout, because diffutils is not installed.

UCRT64 is the documented Windows build path (#36) and ships neither `diffutils` nor `python3`, so this is the default first experience there, not an exotic configuration.

## The change

`cmp` is guarded once and reached through `same()`, which returns `0` / `1` / `2` for identical / differ / no-tool. Call sites that could only say PASS or FAIL now have somewhere to put the third case. `cmp`'s own exit 2 ("could not read a file") folds into 1 — which is what `if cmp -s` already did with it — so **no existing verdict moves**.

The download section already had the right shape: `NO_CURL` skipped all five of its checks. That flag becomes `DL_MISS` and carries the *reason* rather than a bare `1`; `python3` joins `curl` in setting it, and the fixture `python3` builds is guarded behind it too.

**The SIMD check is deliberately left alone.** With no `cmp` it falls through to the fp-noise bound, which is a real verdict on its own and needs no `cmp`. It loses the bit-identity claim and nothing else — so it keeps running rather than skipping.

## Full results

| | before | after |
|---|---|---|
| `cmp` hidden | 36 passed, **8 failed**, 13 skipped | 36 passed, **0 failed**, 21 skipped |
| `python3` hidden | 7 passed, **4 failed**, 19 skipped | 7 passed, **0 failed**, 24 skipped |
| both present | 44 passed, 0 failed, 13 skipped | **unchanged**, exit 0 |

`bash -n` clean. No file outside `tests/run.sh` is touched.

## Two things noted and deliberately not included

Separate concerns; happy to send either as its own PR.

1. **`.gitignore` has no `*.exe` or `*.dll`.** It covers `waste`, `libwaste.a`, `*.so`, `*.dylib`, `sweep` — so a native Windows build leaves `waste.exe`, `sweep.exe`, `libwaste.dll` and `libwastevq.dll` untracked.
2. **A few `python3` calls inside the engine and rotary blocks are still unguarded** (the fp-noise comparators, `make_test_container.py`). Same class of bug — but I did not observe them failing, because by the time this machine had a container-capable run `python3` was installed. This project measures rather than asserts, so I have left out a failure I did not actually see.

## Context

Found while standing up a native Windows x86 measurement box (128 GB, Zen 2, PCIe 4.0 NVMe) to work on Gate 7 and the AVX2 side of #32/#38. This was the first thing in the way. Related to #36, which is where the "gaps no CI job can reach" framing comes from.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -44,6 +44,28 @@ no()   { printf "  \033[31mFAIL\033[0m  %s\n" "$1"; fail=$((fail+1)); }
 sk()   { printf "  \033[33mSKIP\033[0m  %s — %s\n" "$1" "$2"; skip=$((skip+1)); }
 head_() { printf "\n\033[1m%s\033[0m\n" "$1"; }
 
+# Every bit-identity check in this file rests on `cmp`, which is diffutils —
+# and a bare MSYS2/MinGW install ships neither it nor `diff`. Unguarded, an
+# absent cmp made `cmp -s` false and the suite reported "expert cache changes
+# results": the engine called broken when what is missing is a tool, the one
+# thing line 10 says this suite must never do. Same reason curl is guarded
+# below. Measured on a fresh MSYS2 UCRT64 install, where it turned eight
+# passes into eight failures.
+#
+# 0 = identical, 1 = differ, 2 = no tool to tell with — callers must treat
+# 2 as SKIP, never as either verdict.
+HAVE_CMP=1
+command -v cmp >/dev/null 2>&1 || HAVE_CMP=0
+# cmp's own exit 2 means "could not read a file", which the callers below
+# already treated as a difference — so it is folded into 1 rather than left
+# to collide with the sentinel.
+same() {
+    [ "$HAVE_CMP" = 1 ] || return 2
+    cmp -s "$1" "$2" && return 0
+    return 1
+}
+NO_CMP="cmp not installed (diffutils)"
+
 # uv, with a deadline. A uv that is absent skips cleanly; a uv that is
 # present and cannot work does not fail, it *hangs*, and takes the suite
 # with it — MSYS2's mingw-w64-ucrt-x86_64-uv 0.12.1 opened no TCP sockets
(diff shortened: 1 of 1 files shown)


marcobambini:
Merged. Reproduced here first, because the whole point of this change is the
behaviour when a tool is *absent*, which no CI job exercises — so a green run
says nothing about it either way.

Both tools hidden by a PATH holding symlinks to everything else, so the two
sides differ only in the commit. This machine has real Kimi-Linear and K3
containers, so the counts are larger than yours:

**`cmp` hidden**

| | main | this PR |
|---|---|---|
| | 42 passed, **8 failed**, 8 skipped | 42 passed, **0 failed**, 16 skipped |

The eight on main, verbatim: `resume`, `no-range fallback`,
`expert cache changes results`, `read-ahead changes results`,
`router lookahead changes results`, `purgeable slots change results`,
`mla_use_nope: false was read as NoPE and skipped the rotation`,
`rope_scaling null or {} did not load as plain RoPE`. Every one becomes a SKIP
naming diffutils, and the 42 passes do not move.

**`python3` hidden** — the four `range server did not start` failures all become
SKIPs naming the interpreter.

Your reading of why this matters more than a label is right and is why it went
in as it stands: `expert cache changes results` is the report of a bit-identity
violation in the expert cache, and in a log it is indistinguishable from the
real defect. `mla_use_nope` names the precise bug #27 fixed.

## The part you could not observe, observed

You wrote:

> A few `python3` calls inside the engine and rotary blocks are still unguarded
> ... I did not observe them failing, because by the time this machine had a
> container-capable run `python3` was installed.

Leaving out a failure you had not seen was the right call. It also means the
second half was invisible from your machine and visible from this one, which
has both a container and a way to hide the interpreter. With `python3` hidden,
this PR still ends at **13 passed, 11 failed, 24 skipped**:

```
FAIL  chunked prefill diverges
FAIL  quantized storage changes results
FAIL  SIMD backend diverges from the CPU baseline
FAIL  engine diverges from the oracle
FAIL  default budget off the rule (no output)
FAIL  format_version not enforced (rc=1 rc2=1)
FAIL  default budget not capped on K3 (no output)
FAIL  info describes something else (no output)
FAIL  params_total off the rule (no output)
FAIL  info describes something else on K3 (no output)
FAIL  params_total off the rule on K3 (no output)
```

Same defect, different block: the fp-noise comparators and the `waste info` /
`waste plan` parsers are python3, and without it they accuse the engine of
diverging from its own oracle. `(no output)` is the tell — the check never ran.

That is not a criticism of this PR, which does what it says and does it
cleanly. It is the rest of the same contract, and it is now measured rather
than suspected, so it can be closed on purpose. I would rather have it as its
own change than fold it into this one after the fact. Tracking it in #36 with
the reproduction, unless you would like to send it — you found the class.

## On the two you set aside

`.gitignore` has no `*.exe` or `*.dll`: correct, and worth a separate one-liner.

53 passed / 0 failed / 6 skipped here with everything installed, unchanged from
the merge-base. Thank you — this is the report I should have acted on faster,
and the one where being left at `action_required` cost the most.

PR #46, opened 2026-08-22
Prefer experts the cache already has, for ranking only

mfethe1:
**Stacked on #45.** I would have targeted that branch directly, but GitHub will not accept a base branch that lives on a fork, so this targets `main` and therefore **carries #45's commit as well as its own**:

- `c4afbde` — `WASTE_DUMP_SCORES`, which is **#45**, review it there
- `33b4d2e` — this PR's actual change

**Read only the second commit's diff here.** If #45 merges first, this rebases to just `33b4d2e` and I will force-push it clean.

**Read this first, because it is the reason this is a knob and not a proposal to change anything:** this does **not** reproduce the greedy continuation the way `num_experts_per_token=8` does. README accepted top-8 against *two* bars — KL 0.037 **and** "it reproduces top-16's greedy continuation on the prompts tested". This clears the KL bar at a smaller value and stays on-task everywhere I looked, but the text is not byte-identical. That asymmetry is the whole reason it is offered off by default, and I would rather lead with it than have you find it.

## What it does

Experts already in the cache get a bonus **for ranking only**, per [arXiv:2412.00099](https://arxiv.org/abs/2412.00099):

```c
v = score[e] + bias[e] + lambda * layer_logit_range * resident[e]
```

`w[j]` still takes the untouched `score[best]`, so the gating weights are exactly what the model computed and only the *choice of which K to run* moves. Off unless `WASTE_CCR_LAMBDA` is set; unset, `make check` is 44 passed / 0 failed / 13 skipped, unchanged.

## Measured, Kimi-Linear, lookahead ON

| lambda | hit% | GB/token | bytes saved | KL from lambda=0 |
|---|---|---|---|---|
| 0 | 70.6 | 0.322 | — | 0 |
| 0.05 | 76.9 | 0.294 | 8.7% | 0.0166 |
| **0.10** | **82.5** | **0.249** | **22.7%** | **0.0232** |
| 0.25 | 87.0 | 0.227 | 29.5% | 0.0525 |
| 0.50 | 88.2 | 0.224 | 30.4% | 0.0670 |

Against README's own prices: **top-8 is KL 0.037 and ships; top-4 is KL 0.118 and was refused.** lambda=0.1 sits below the accepted one.

### Evidence

| Before (lambda=0) | After (lambda=0.1) |
| --- | --- |
| 70.6% hit, 0.323 GB/token | 82.5% hit, 0.249 GB/token |

`tok/s` is filtered out of both frames on purpose — an unrelated download was writing to the same disk throughout. Hit rate and bytes are deterministic counters; throughput on this box is not, and I would rather show you nothing than show you a number I cannot stand behind.

## Why I believe it is additive to the lookahead rather than the same gain twice

That was the obvious way for this to be worthless: a prefetcher that makes about-to-be-needed experts resident, and a prior that prefers resident experts, could easily be chasing the same misses. §35 prices the lookahead at 14–19% → 38–40% demand hit rate on its own.

**`lambda=0` reproduces the engine's independently measured 70.6% baseline exactly**, with the lookahead on in every arm. The +11.9 points at 0.10 are on top of it.

An offline simulator I wrote first said +20 points at the same lambda. It was wrong by ~40% because it modelled no lookahead — its baseline sat 26 points under the engine's. Mentioning it because it is the kind of over-estimate this idea invites.

## The §54 check

`docs/LEARNED.md` §54 is the reason I did not stop at a hit-rate number: `k3-k16` had **100% hit, zero misses** and ran at **half** the speed, because `kda` and `mla` — which touch no experts — went 1.6x slower from ~45 GB being permanently hot.

First profile run here looked exactly like that in reverse: everything sped up ~1.5x including `kda`. **Interleaved repeats showed it was slow-first-run noise** — by rep2 the two arms were within 1.8% and `kda` was 3.39s vs 3.42s. **No machine-wide effect.** Plausibly because this only reduces misses inside an already-bounded 1.65 GB cache, rather than making an entire working set permanently resident.

## Quality, beyond KL

Greedy `--temp 0`, three prompt types, lambda 0 / 0.05 / 0.1 / 0.25:

- **Factual recall** (planets, strongest magnetic field): every lambda gives the correct order, answers Jupiter, and gives the correct mechanism. The numeric details wobble — but the *baseline* conflates surface field strength with magnetic moment, so the variants shuffle an imprecision that was already there.
- **Technical explanation**: all accurate and on-task. lambda=0.25 is arguably more precise than baseline.
- **Code** (longest strictly-increasing run), graded by **executing** it against 7 cases rather than reading its docstring: **7/7 correct at every lambda**, including 0.25. Worth noting the baseline's own example comment is wrong (`[10,9,2,5,7,8] -> 2`; it is 4) while lambda=0.25 got both of its right — comment arithmetic is this model's baseline behaviour, not something the prior introduces.

**No compounding.** I expected a state-dependent perturbation (this one depends on cache contents, which shift as generation proceeds) might accumulate where top-k's fixed perturbation does not. Across 160- and 300-token generations, late text is as clean as early text at every lambda. Recorded as unsupported rather than refuted — two lengths on one model is not a proof.

## What this does not establish

- **Kimi-Linear only.** K3 is top-16-of-896 over a ~950 GB expert set — a completely different residency regime, where the cache holds a far smaller fraction. The signal could be weaker (less worth preferring) or stronger (more tail to avoid). §54's k3-k16 is the standing warning against assuming transfer, and I have no K3 container yet (downloading).
- **KL is one position on one prompt.** §25 is the standing warning about a number taken once.
- **The continuation bar is not cleared**, as above.
- One deviation from the paper: it scales by the layer's *instantaneous* logit range rather than a running mean, because static mutable state would be wrong when `waste.h` promises several models open in one process.

Happy to drop it, restructure it as a manifest field rather than an env var, or hold it until there is a K3 number to pair with it.

## Context

Native Windows x86 measurement box (Zen 2, 128 GB, PCIe 4.0 NVMe) stood up for #37/#38. Companions: #42, #43, #44, #45.

### DIFF
--- a/src/ecache.c
+++ b/src/ecache.c
@@ -593,6 +593,23 @@ void waste_ecache_hint(waste_ecache *c, int layer, const int *ids, int n)
     ec_unlock(c);
 }
 
+void waste_ecache_resident_mask(waste_ecache *c, int layer, int n, uint8_t *out)
+{
+    if (!out || n <= 0) return;
+    memset(out, 0, (size_t)n);
+    if (!c || c->n_slots <= 0) return;
+
+    ec_lock(c);
+    for (int e = 0; e < n; e++) {
+        const int si = ec_lookup(c, ec_key(layer, e));
+        /* READY only. An INFLIGHT slot has been asked for and has not
+         * arrived, so a router told it was resident would pick it and then
+         * wait on the same read it was trying to avoid. */
+        if (si >= 0 && c->slot[si].state == EC_READY) out[e] = 1u;
+    }
+    ec_unlock(c);
+}
+
 const uint8_t *waste_ecache_get(waste_ecache *c, int layer, int expert,
                                 waste_fetch_fn fetch, void *user)
 {
--- a/src/ecache.h
+++ b/src/ecache.h
@@ -167,6 +167,22 @@ void waste_ecache_io_stop(waste_ecache *c);
  * Reads for the first `depth` of them that are not resident start now, and
  * each waste_ecache_get releases one more into the pipe. Calling it with a
  * synchronous cache is a no-op, so call sites need no conditional. */
+/* Residency of a whole layer's experts, without touching any of them.
+ *
+ * `out[e]` becomes 1 when expert e of `layer` is in the cache and READY —
+ * usable with no disk read. Nothing is claimed, no recency moves, and
+ * nothing counts as a hit or a miss: this answers a question, it does not
+ * make a request. INFLIGHT deliberately reads as 0, so "resident" means
+ * arrived rather than asked for.
+ *
+ * One call per layer rather than one per expert because a single mutex
+ * covers the whole cache, and a per-expert probe would take it E times per
+ * layer per token — 83k times a token on K3.
+ *
+ * For a router that wants to prefer experts it already has (arXiv:2412.00099).
+ */
+void waste_ecache_resident_mask(waste_ecache *c, int layer, int n, uint8_t *out);
+
 void waste_ecache_hint(waste_ecache *c, int layer, const int *ids, int n);
 
 /* Speculative fill for a layer that has not routed yet. Unlike a hint these
--- a/src/model.c
+++ b/src/model.c
@@ -164,6 +164,7 @@ static int q8_off = 1;     /* 1 = keep the trunk stored as int8          */
 static int sdot_on = 0;    /* 1 = also quantize activations (SDOT path)  */
 static int i8mm_on = 0;    /* SMMLA batched matmul; costs activation int8 */
 static const char *dump_route = NULL;  /* WASTE_DUMP_ROUTE, see moe_layer */
+static const char *dump_scores = NULL; /* WASTE_DUMP_SCORES, see moe_layer */
 /* Absolute position of the first token of the pass being routed. The dump
  * names each row by the token it belongs to rather than leaving a reader
  * to infer it from where the layer index wraps — which is a heuristic
@@ -193,6 +194,7 @@ static void model_opts_init(void)
     /* Read once rather than per layer per token: moe_layer runs 92
      * times a token and getenv is not free. */
     dump_route = getenv("WASTE_DUMP_ROUTE");
+    dump_scores = getenv("WASTE_DUMP_SCORES");
     /* How many of the next layer's experts to fetch on the router's guess.
      * The layer boundary holds about six reads and the prediction's
      * precision falls off past there, so that is the default. 0 is off. */
@@ -2745,6 +2747,32 @@ static void moe_layer(waste_model *m, int L, const float *in, float *out, int *r
     }
     for (int j = 0; j < K; j++) w[j] *= c->routed_scale;
     if (routed) for (int j = 0; j < K; j++) routed[j] = idx[j];
+
+    /* WASTE_DUMP_SCORES=path appends one line per (token, layer):
+     *
+     *     pos  L  v0 v1 .. v(E-1)
+     *
+     * the *selection* value for every expert, `score[e] + bias[e]` — the
+     * quantity the loop above ranks on, not the weight it later applies.
+     *
+     * WASTE_DUMP_ROUTE records the ids that won, and no question about a
+     * *different* ranking can be answered from winners alone. Anything that
+     * would reorder the top-K — a residency prior, a dynamic k, a pruning
+     * criterion — needs to know what the experts that lost were worth, and
+     * by how much they lost. A trace of the chosen cannot say.
+     *
+     * Same reasoning as the comment below, and the same economics: this is
+     * one fprintf against a rebuild plus a sweep. */
+    if (dump_scores) {
+        FILE *sf = fopen(dump_scores, "a");
+        if (sf) {
+            fprintf(sf, "%d %d", dump_pos0, L);
+            for (int e = 0; e < E; e++)
+                fprintf(sf, " %.6g", score[e] + (bias ? bias[e] : 0.0f));
+            fputc('\n', sf);
+            fclose(sf);
+        }
+    }
 
     /* WASTE_DUMP_ROUTE=path appends one line per (token, layer):
      *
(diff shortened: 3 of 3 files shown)


mfethe1:
Follow-up: I widened the quality evidence, because the continuation claim is the part of this PR a reviewer has to take most on trust and the original battery only tested *coherence*. This tests whether **constraints survive** — the behaviours a perturbed router should break first.

Five prompts, `--temp 0`, λ ∈ {0, 0.1, 0.25}, graded by parsing rather than by reading:

| test | λ=0 | λ=0.1 | λ=0.25 |
|---|---|---|---|
| time arithmetic (`14:20 + 3h55m + 40m delay`) | PASS | PASS | PASS |
| exactly three bullets, no preamble or closing | PASS | PASS | PASS |
| `17 × 23` with working shown | PASS | PASS | PASS |
| bare valid JSON, three named keys, no fence | PASS | PASS | PASS |
| exactly five paragraphs | PASS | PASS | PASS |

**Nothing breaks at λ=0.1 or 0.25 that does not also break at λ=0.** The two strict-format cases are the ones I would weight: the JSON prompt produced bare parseable output with the right keys and no markdown fence at every λ, and the three-bullet prompt produced exactly three bullets with nothing before or after at every λ.

That is on top of what was already in the PR body — three coherence prompts, and the code test graded by executing the function against 7 cases (7/7 at every λ).

### Two of my five initial "failures" were bugs in my grader, not the model

Recording these because they cut the other way, and because the same shape of mistake is what a reviewer should suspect me of:

- **Time arithmetic**: I put `19:15` in the grader as the expected answer. It is `18:55` — `14:20 + 3h55m = 18:15`, plus the 40-minute delay. The model produced `18:55` at every λ with identical reasoning. My arithmetic was wrong, not its.
- **Five paragraphs**: at 220 tokens both arms truncated mid-sentence, so "2 paragraphs" was measuring my token budget. Re-run at 600 it is five paragraphs at every λ.

Three of the four failure signals across both batteries have turned out to be artefacts of the harness rather than the change under test. It is the reason every arm here has a λ=0 control — a failure that reproduces at λ=0 is the model, not the routing.

### The caveat in the PR body is unchanged

This still does **not** reproduce the greedy continuation byte-for-byte. README accepted `num_experts_per_token=8` against *two* bars, and this clears one of them. Wider constraint testing does not change that; it only says the divergence is paraphrase rather than drift, across every category I could think to check.

Still Kimi-Linear only. No K3 number yet — the container is downloading, and §54's `k3-k16` is the standing reason not to assume it transfers.

marcobambini:
Merged, as an off-by-default knob and on the strength of you leading with what
it does not do.

**Reproduced here** — M5 Pro, Kimi-Linear VQ3R, `--budget 4G`, `-n 120`,
lookahead on in every arm:

| lambda | hit | GB/token | bytes saved |
|---|---|---|---|
| unset | 76.1% | 0.231 | — |
| 0.10 | **89.5%** | 0.167 | **27.7%** |
| 0.25 | 92.8% | 0.143 | 38.1% |

Different budget than yours, so the absolute hit rates differ; the direction
and the magnitude land in the same place. `WASTE_CCR_LAMBDA=0` is byte-identical
to the merge-base, and so is a run with the variable unset.

**And the continuation does diverge**, which I checked because you said it
would and because it is the bar that decides how this ships. Greedy, `--temp 0`,
one code prompt:

```
lambda unset : A "run" is a contiguous subsequence where each element is ...
lambda 0.10  : A "run" is a consecutive subsequence where each element is ...
```

On-task, correct, and not the same text. That is the asymmetry you led with,
confirmed on a second machine and a second prompt.

**So the knob is in and the default is not moving**, and I want the reason on
the record rather than implied. README prices top-8 against *two* bars — KL
0.037 **and** it reproduces top-16's greedy continuation. You clear the first
at a smaller value and not the second. A knob that is off unless asked for is
a different proposition from a default, and the measurement is worth having in
the tree where people can run it; changing what every user gets with no flag
needs the second bar too, or a deliberate decision to retire it.

**One number to be careful with.** tok/s here went 9.97 → 9.86 → 9.76 as bytes
fell 38%. Not a regression worth worrying about — it is a warm cache on a fast
internal SSD, where disk is not the binding term, so removing reads buys
nothing and the extra ranking work is visible instead. It does mean the byte
saving and a throughput win are not the same claim on every host, and you were
right to filter tok/s out of your own frames rather than show a number you
could not stand behind.

Your §54 check is the part that made this reviewable. `k3-k16` had 100% hit and
ran at half speed, so a hit-rate improvement is not self-evidently a win, and
"first profile looked exactly like that in reverse, then interleaved repeats
showed it was slow-first-run noise" is the check most people would have skipped
after seeing the answer they wanted. #37 and #44 have both since reported the
same first-run artefact on other hosts, so it was real and it was general.

K3 remains the open question, as you say — top-16-of-896 over a ~950 GB set is a
different residency regime and §54 is the standing warning against assuming
transfer. If you get a K3 container, that arm is the one worth running.

mfethe1:
**K3, both requested arms — top-16-of-896 over the ~950 GB set.**

Windows 11, Ryzen 7 3700X, `test_forward`/`waste bench` on a real K3 container (982 GB, unmodified 4-bit trunk + VQ3R experts, byte-for-byte integrity-verified before and after each run). Fresh process per arm. Two separate measurements, same direction both times:

**Run 1** — `--budget 32G` (3.17 GB expert cache), `-n 10`, commit `878352b` (v0.6.9):

| lambda | hit | GB/token |
|---|---|---|
| unset | 16.9% | 32.069 |
| 0.10 | 17.4% | 31.920 |
| 0.25 | 17.7% | 31.811 |

**Run 2** — `--budget 37G` (8.17 GB expert cache), `-n 20`, commit `0e796c3` (`src/` functionally identical to `878352b` — only `waste.h`'s version macros differ, confirmed via `git diff --stat`):

| lambda | hit | GB/token |
|---|---|---|
| unset | 23.7% | 23.979 |
| 0.10 | 32.4% | 21.522 |
| 0.25 | 38.2% | 20.170 |

Same direction (hit up, bytes down as lambda rises), same monotonic ordering across arms, both real hardware. The magnitude is where it gets interesting: run 1's baseline→0.25 delta is +0.8pp hit / −0.8% bytes; run 2's is +14.5pp hit / −15.9% bytes — roughly 18–20x the effect at ~2.6x the cache. That's consistent with CCR's eviction bias only having room to matter once cache size makes eviction genuinely contested — at 3.17 GB (a sliver of the ~950 GB expert set) there's barely any contested eviction to bias; at 8.17 GB there evidently is much more. Two points isn't a scaling curve, so I'm not asserting a functional form, but the residency regime you flagged does look like the load-bearing variable here, not container size alone.

tok/s is noise-dominated at these n (10–20 tokens) per this repo's own §54 discipline, so I'm not reporting it — hit%/GB-per-token are the trustworthy numbers.

PR #53, opened 2026-08-27
Bank striping across N devices: WASTE_BANK_SHARDS, record-granularity round-robin

mfethe1:
## What

Optional multi-device expert-bank striping. Set `WASTE_BANK_SHARDS=/mnt/a,/mnt/b` and each layer's bank opens as N shard files (bank basename in each dir); expert `e` is read from shard `e % N` at offset `(e / N) * rec_bytes`. `tools/split_banks.py` produces and byte-verifies shard sets. Unset env = exact current behavior — unstriped is the N=1 case of the same code path.

## Why

Per-token demand is top-k experts of a *single* layer — one file today, so one device serves every expert read of a token. Round-robin record placement spreads a token's reads across devices, so effective bandwidth scales toward min(N, readers) with the existing async reader pool (`WASTE_IO_THREADS`). On a 2-drive rig this targets ~1.8-2x; on 4 NVMe with more readers, more.

No container format change: shards are plain files named after each bank. Fail-closed: any missing or short shard refuses the load.

## Evidence

- tiny.waste split across 2 dirs: logits byte-identical to unstriped (sha256 `de62689a…`, argmax + max-logit equal) — correctness gate is identity, not tolerance
- `tools/split_banks.py --mode verify`: every record byte-identical post-split
- `make check` 48 passed / 0 failed (13 skips: no real container locally)
- ASan+UBSan striped run: 0 findings, same-build striped==unstriped byte-exact

## Not claimed

No device A/B numbers yet — needs a 2-drive rig with a real container. This PR is the mechanism; the bandwidth measurement follows on hardware.

### DIFF
--- a/src/model.c
+++ b/src/model.c
@@ -1067,7 +1067,10 @@ int waste_model_load(waste_model *m, const char *dir, int kv_cap,
     memset(m, 0, sizeof *m);
     pthread_mutex_init(&m->fetch_mu, NULL);
     m->trunk_fd = -1;
-    for (int L = 0; L < WASTE_MAX_LAYERS; L++) m->bank[L].fd = -1;
+    for (int L = 0; L < WASTE_MAX_LAYERS; L++) {
+        for (int s = 0; s < WASTE_MAX_SHARDS; s++) m->bank[L].fd[s] = -1;
+        m->bank[L].n_shards = 1;
+    }
     m->want_vision = opt->want_vision;
     m->want_direct = opt->direct_io;
     pthread_once(&model_opts_once, model_opts_init);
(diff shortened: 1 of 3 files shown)


mfethe1:
Repointed the PR branch to a clean single-commit base on current `main` (878352b5) — it was previously stacked on the unmerged #48 branch, which made the diff carry unrelated commits and left `mergeable_state: dirty`.

Verified on the rebased head before pushing: `make check` 44 passed / 0 failed / 13 skipped; split_banks.py split+verify roundtrip on a fresh `make_test_container` fixture (all records byte-identical across shards); logits byte-identical striped vs unstriped (`--json` sha256 match on both arms); ASan/UBSan clean in the local suite. CI is now 9/9 green on `bc710d5` and the diff is exactly the three intended files (`src/model.c`, `src/model.h`, `tools/split_banks.py`). The striping behavior is unchanged from the original description — only the base and the missing SPDX header in the new tool are different.

Issue #10, opened 2026-08-01
x86: SIGFPE (divide-by-zero) in waste_model_load when container carries skipped vision tensors — info/run crash, plan works


jenbrannstrom:
## Summary

On x86, `waste info` and `waste run` crash with an integer divide-by-zero in `waste_model_load` when the container carries tensors that the loader skips (vision tower / projector). `waste plan` works. The bug is architecture-dependent undefined behavior, not a Windows issue: AArch64 `sdiv` silently returns 0 for division by zero, so Apple-silicon runs never trap; x86 `idiv` raises `#DE`. Any x86 build (Linux included) loading a container with vision sidecars should hit it.

## Environment

- WASTE 0.6.0, commit `ce96e38`, AVX-512 backend
- Windows 11 x64, built with MinGW-w64 (gcc 16.1.0, UCRT/POSIX): `mingw32-make -j8 CC=gcc ARCH=x86_64-w64-mingw32 WINDOWS=1`
- Container: full Kimi-K3 converted from the official HF weights with `tools/convert.py` (92/92 expert banks, ~981 GB), text-only run (no `--image`)
- 62 GB RAM machine, container on internal NVMe

## Symptom

```
> waste plan C:\models\k3.waste     # OK, prints the memory plan
> waste info C:\models\k3.waste     # exits immediately
exit code -1073741676 (0xC0000094, STATUS_INTEGER_DIVIDE_BY_ZERO)
> waste run C:\models\k3.waste "Hello" -n 2    # same crash
```

## Backtrace

```
Thread 1 received signal SIGFPE, Arithmetic exception.
0x00007ff6ef37bce0 in waste_model_load ()
#0  waste_model_load ()
#1  waste_open ()
#2  open_model ()
#3  cmd_info ()
```

Faulting instruction is `idiv %r9d` with `r9d = 0`; the surrounding code loads `shape[ndim-1]` (7168 here) and does the `(x + g - 1) / g` round-up.

## Root cause

In the trunk-tensor loop in `src/model.c` (`waste_model_load`), non-text tensors (`vision_tower.*` / `mm_projector.*` when vision isn't requested, and anything outside `cfg.prefix`) take the early `continue` that sets `t->on_disk = 1` — **before** `t->group` is ever assigned (that only happens in the kept-quantized branch). The tensor array is `calloc`'d, so those tensors keep `group == 0`.

The row-scratch sizing block later iterates **every** on-disk tensor:

```c
[...]

PR #63, opened 2026-09-04
Added preliminary support for Qwen3.8-flash-next

Skibisky:
Qwen3.8-Flash-Next is a 125B model with 6B active experts plus 51B n-gram embedding and 4B MTP.

Unable to hold it all in memory at once, I decided to attempt to fork WARP to make use of the streaming, cache, and similar technologies for both the MoE and the n-gram embeddings. 

An attempt was made to incorporate the MTP for token prediction but on the hardware available the accuracy and time taken never was able to deliver a gain due to contention of shared resources (this work will be provided on a future branch on my own fork as a reference).

----
This adds text-only support for Qwen3.8-Flash-Next’s qwen4_exp architecture. The converter handles its nested Hugging Face configuration, packed expert tensors, and 128 PLE embedding shards while reusing WARP’s existing prefix, rename, quantization, streaming, and resume machinery. Tokenizer support also gains Qwen’s single-digit pre-tokenization behavior without introducing an architecture-specific tokenizer path.

The CPU runtime implements Qwen’s Gated DeltaNet and sparse-attention layers, HyperConnection residual streams, PLE retrieval, and routed MoE with a gated shared expert. These components integrate with WARP’s existing expert cache, dynamic expert-parallel execution, memory planning, state persistence, and container format. Unsupported capabilities—including vision, video, MTP, and native Qwen serving—are explicitly refused or left outside this PR’s scope.

Validation includes standalone C kernels checked against independent PyTorch references, a synthetic qwen4_exp container, converter and round-trip coverage, tokenizer tests, and a container-native forward-pass oracle. Routes match exactly, the worst hidden-state difference is 2.4e-7, and final logits differ by 1.9e-6 with the same argmax. On the converted 124 GB checkpoint, the runtime produces coherent output and sustains approximately 7 tok/s using eight threads and an 8 GiB expert cache

----

The heavy lifting was performed by Codex Sol 5.6, occasionally using Luna sub-agents, and the occasional Claude Opus review.

Additional on-going investigations were trying to 'outsource' non-critical calculations with low memory contentions to Metal/GPU, which isn't looking so great.

I am posting up this work and submitting this PR because [Qwen3.8-Max](https://huggingface.co/Qwen/Qwen3.8-2.4T-A95B) is out and might share similar architecture.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,8 @@ CFLAGS  += -MMD -MP
 
 SRC := src/model.c src/kda.c src/backend.c src/ecache.c src/version.c \
        src/tokenizer.c src/waste.c src/vq.c src/vision.c src/image.c \
-       src/crc32.c src/memory.c
+       src/crc32.c src/memory.c src/qwen_gdn.c src/qwen_qsa.c \
+       src/qwen_hc.c src/qwen_ple.c src/qwen_moe.c
 # Match what backend.c tests for. Linux/aarch64 reports "aarch64", which
 # does not contain "arm" — the old findstring left kda_neon.c out of the
 # build while backend.c still emitted the call to it, so the link failed
@@ -258,8 +259,8 @@ waste$(EXE): cli/main.o libwaste.a
 # the two failures tests/run.sh was written to catch, so a binary that
 # `test` builds and `clean` forgets defeats the check meant to notice it.
 TESTNAMES := test_kda test_container test_forward test_tokenizer test_k3parts \
-             test_state test_vision test_vision_glm test_image test_memory \
-             test_cpus test_lock sweep
+             test_qwenparts test_state test_vision test_vision_glm test_image \
+             test_memory test_cpus test_lock sweep
 TESTBINS  := $(addsuffix $(EXE),$(TESTNAMES))
 
 test: $(TESTBINS)
@@ -286,6 +287,8 @@ sweep$(EXE): tests/sweep.o libwaste.a
 test_tokenizer$(EXE): tests/test_tokenizer.o libwaste.a
 	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
 test_k3parts$(EXE): tests/test_k3parts.o libwaste.a
+	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
+test_qwenparts$(EXE): tests/test_qwenparts.o libwaste.a
 	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
 test_image$(EXE): tests/test_image.o libwaste.a
 	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
--- a/docs/LEARNED.md
+++ b/docs/LEARNED.md
@@ -5412,3 +5412,57 @@ nothing at all — silently, since the conversation still came out valid.
 is the other half. **A test that only compares a thing to itself is not a
 weak oracle, it is not an oracle**, and the three protocols this repo
 renders now each have one.
+
+## 74. Qwen3.8-Flash-Next: what was new, and what only looked new (2026-09-04)
+
+Four architectural pieces this engine had never run — Gated DeltaNet,
+Qwen Sparse Attention, HyperConnection, and a per-layer n-gram embedding —
+and the useful finding is how little of the *engine* they touched. The
+container format did not change: a packed `[E, 2I, H]` gate_up beside an
+`[E, H, I]` down splits into exactly the WEXP records everything else
+writes, one expert per 4 KiB-aligned record, so routing still costs one
+`pread` and the expert cache, the read-ahead and the expert-parallel path
+were reused unmodified. What is genuinely Qwen's is five self-contained
+kernel files and a loader branch.
+
+**The 80 GiB trunk that is 2.6 GB resident.** The n-gram tables are 16
+heads of ~20 M rows, 78 of the trunk file's 80.51 GiB. Held resident they
+would exceed the whole RAM budget on any machine this targets; read a row
+per head per token they cost one Q8G row each. The same exclusion the
+embedding table has always had, for the same reason, and it is what makes
+a 176.94 B model open with a 3.11 GB floor. The converter has the mirror
+problem: a head is ~12 GiB as f32, so it is written 64 Ki rows at a time,
+relying on Q8G grouping along the last dimension to make the batches
+reconstruct what quantizing the head whole would have produced.
+
+**8 GiB of expert cache, and nothing above it.** Measured over 48 greedy
+tokens: 4 GiB gives 3.20 tok/s at an 8% hit rate, 8 GiB gives 4.97 at 64%,
+16 GiB gives 4.92 at 88%. The collapse below one working set is §3's rule
+again — below a multiple the hit rate is zero, not low. The flat top is
+the more useful half: **24 points of hit rate bought nothing**, because at
+64% the remaining reads already overlap the arithmetic. A cache sized to
+the machine rather than to the knee spends RAM for no tokens.
+
+**A tokenizer difference that no vocabulary test could see.** Qwen's
+pre-tokenization pattern is `\p{N}` where Kimi's and GLM's are
+`\p{N}{1,3}`: every digit is its own piece. `tools/hf_tokenizer.py` was
+right to refuse the pattern rather than approximate it, and the engine now
+carries `tokenizer_digit_run` the way it already carried
+`tokenizer_han_split`. The trap is in the checking, not the fixing —
+**Qwen's vocabulary contains no multi-digit token at all**, so on this
+checkpoint the two settings produce identical ids and every parity test
+passes either way. "202" has no merge to reach. The flag is therefore
+tested on a synthetic vocabulary that does hold `20`, where the pre-token
+boundary is directly visible. A parity test against the release would have
+green-lit the wrong pattern for the next member of the family.
+
+**An unexplained 4e-3.** The container-native oracle and the engine read
+the same trunk dequantized to the same f32, so their difference should be
+summation order — around 1e-6. It is 4e-3 relative per layer. Routed
+expert ids and weights match exactly at every layer and the argmax matches,
+so nothing observable is wrong, and end-to-end generation is coherent on
+the real checkpoint. It is recorded here undiagnosed rather than absorbed
+into a tolerance: 4e-3 is close to bf16's epsilon and nothing in that path
+should be rounding to bf16. The suite gates at the measured value so the
+number cannot grow while nobody is looking, which is the least a check can
+do about a thing it does not understand.
--- a/docs/QWEN.md
+++ b/docs/QWEN.md
@@ -1,0 +1,181 @@
+# Qwen3.8-Flash-Next
+
+Text inference for `qwen4_exp` / `qwen4_exp_text`. Four architectural
+pieces that no other model here has, and a container format that does not
+change to hold them.
+
+Pinned for every measurement below:
+
+    Qwen/Qwen3.8-Flash-Next
+    revision de4b8e4d43b917e7706784d8bb445c9af86a3540
+
+131 shards, 360,013,002,208 bytes on the hub, index SHA-256
+`99e815241ef03325536b0aaa4441deea45174c17fae31e10f0bb456410c590de`.
+Fetch the revision, not `main` — this repository's `main` has moved since,
+and a container built from a different one is not the container these
+numbers describe.
+
+## What is different about it
+
+| | Kimi / GLM | Qwen3.8-Flash-Next |
+|---|---|---|
+| recurrent layers | Kimi Delta Attention | **Gated DeltaNet** — per-head scalar decay, 16 QK heads repeated onto 48 V heads |
+| attention layers | MLA over a compressed latent | **Qwen Sparse Attention** — an indexer scores 4-key mean-pooled blocks, keeps the best 512 plus the token tail, and attention runs over the *original* K/V |
+| residual | one stream (four on GLM's mHC) | **HyperConnection** — four streams mixed through a rank-320 bottleneck and recombined through a per-branch gate |
+| extra embedding | none | **PLE** — an n-gram lookup at one layer, 16 hashed head tables of ~20 M rows each |
+| router | sigmoid plus a learned bias | **softmax over all 512, then top-k** |
+| experts on disk | one tensor per expert per matrix | **two packed tensors per layer** |
+
+Layer mix is 36 `linear_attention` and 12 `full_attention`, stated by
+`layer_types` and read from it. A container that does not say is refused:
+"every layer is GDN" is a plausible-looking default that would be wrong in
+every token and visible in none.
+
+Shape: hidden 2560, 48 layers, vocab 248320, 512 experts with top-10
+routed plus one gated shared expert of width 640, GDN 16×128 QK and 48×128
+V, QSA 24 query and 2 KV heads at dim 256 with 64 rotary dims, indexer MQA
+at dim 128 with a 2048 budget, `hc_count` 4 / `hc_lowrank` 320, n-gram
+size 3 over 16 heads, `ple_layer_ids: [2]`.
+
+`ple_layer_ids` is 1-based and the PLE tensors live on
+`model.language_model.layers.1`. Two spellings of the same layer; the
+container carries the 0-based one.
+
+## Converting
+
+```bash
+tools/fetch_weights.sh --repo Qwen/Qwen3.8-Flash-Next \
+    --revision de4b8e4d43b917e7706784d8bb445c9af86a3540 \
+    --dest /path/to/raw
+
+uv run --with torch --no-project python tools/convert.py \
+    --src /path/to/raw --out /path/to/qwen38-flash-next.waste
+```
+
+No Qwen-specific flags. Three things happen that do not happen for any
+other family:
+
+* **Packed experts.** `experts.gate_up_proj` [E, 2I, H] and
+  `experts.down_proj` [E, H, I] hold a whole layer's routed experts. They
+  are split into ordinary WEXP records — one expert per 4 KiB-aligned
+  record, format v0, nothing about the read path changes. The layout is
+  validated against its invariant rather than against Flash-Next's
+  dimensions, so a larger family member with the same packing converts on
+  the same code.
+
+* **PLE.** 128 source shards become 16 Q8G head tensors on the trunk. A
+  head is ~12 GiB as f32, so it is quantized 64 Ki rows at a time with one
+  source shard resident; Q8G groups along the last dimension, which is
+  what makes the concatenation of the batches equal to quantizing the head
+  whole. The i64 offset and vocabulary-size tables go into the manifest as
+  integers — they are primes near 2×10⁷ and do not survive a float.
+
+* **`--jobs` defaults to 1.** A worker holds a whole layer's packed pair,
+  3.3 GiB of BF16 before the f32 it dequantizes into. Three at once is
+  what turns a conversion into a swap storm. An explicit `--jobs` wins.
+
+`--reclaim` works, with the n-gram shards as a consumer of their own:
+`build_ple` runs after the trunk pass, so a shard holding both an expert
+and an n-gram slice is not released until the 16 heads are on the trunk.
+The vision tower and the MTP layer have no consumer at all here and are
+released first.
+
+`tools/verify_container.py` reads packed sources and checks PLE rows
+against their source shard, three rows per head, without dequantizing a
+head.
+
+## Running
+
+```bash
+./waste run /path/to/qwen38-flash-next.waste "The capital of France is" \
+    --budget 8G
+```
+
+Measured on an Apple silicon laptop, 12 logical CPUs (8 performance, 4
+efficiency), 48 GiB of RAM, container on the internal SSD, at the commit
+that ships this file:
+
+| | |
+|---|---:|
+| container | 123.9 GiB |
+| ├ `trunk.bin` | 80.51 GiB |
+| ├ expert banks | 42.5 GiB |
+| resident trunk | 2.60 GB |
+| memory floor | 3.11 GB |
+| one token's expert working set | 888.7 MB |
+| parameters | 176.94 B total, 57.87 B active per token |
+
+The gap between an 80.51 GiB trunk file and a 2.60 GB resident trunk is
+the point of the PLE design: the 16 n-gram heads are 78 GiB of the file
+and are read one row per head per token, never held. `waste_plan_memory`
+excludes them from the resident set for the same reason it excludes the
+embedding table.
+
+Throughput, 48 tokens greedy from the same prompt:
+
+| expert cache | tok/s | hit rate |
+|---:|---:|---:|
+| 4 GiB | 3.20 | 8% |
+| 8 GiB | 4.97 | 64% |
+| 16 GiB | 4.92 | 88% |
+
+**8 GiB is the knee and there is nothing above it.** Below one token's
+working set the hit rate collapses rather than degrades — the 4 GiB row is
+that, not a gentle slope — and above 8 GiB a better hit rate buys no time
+at all: the reads it saves were already overlapping the arithmetic. That
+is the whole reason the default budget is not "as much as the machine
+has".
+
+Threads, at 8 GiB: 5.04 tok/s at the default (one per logical CPU), 5.38
+at `--threads 8`, 5.14 at 6, 5.09 at 12. Eight — the performance-core
+count on this machine — is worth about 7%, which is real but is a property
+of this machine and not of the architecture, so it is left to `--threads`
+rather than compiled in as a default. LEARNED §47 has the same finding
+inverting between two other models.
+
+Routed experts go through the existing expert-parallel path and its
+existing per-layer decision: one task per expert when the records are
+already resident, one task per row range when holding a batch would
+barrier the read-ahead. `WASTE_XPAR=0/1` still forces it.
+
+## Correctness
+
+* **Components.** `tests/test_qwenparts.c` dumps every intermediate of the
+  five kernels and `tools/qwenparts_ref.py` recomputes them in PyTorch
+  from the published equations — at the official geometry as well as at
+  toy sizes. Largest disagreement across the dump: 2.4e-7 absolute.
+
+* **Whole forward pass.** `tools/qwen_container_ref.py` implements the
+  same forward pass in PyTorch reading the same container, so the
+  comparison is against an independent decode of identical weights. On the
+  synthetic fixture the routed expert ids and weights match **exactly** at
+  every layer and the logits argmax matches. The worst hidden-state
+  difference is 2.4e-7 absolute and the final-logit difference is 1.9e-6.
+
+* **Tokenizer.** Qwen's pre-tokenization pattern splits every digit into
+  its own piece where Kimi's and GLM's take up to three. The engine is
+  told which through `tokenizer_digit_run`; `tools/hf_tokenizer.py`
+  refuses a pattern it does not recognise rather than approximating one.
+  Against the release's own tokenizer the C encoder matches on every
+  string in `tests/test_qwen_tok.py`, numbers included. The rank file this
+  path writes is byte-identical to one built from `vocab.json` through the
+  GPT-2 byte map.
+
+* **End to end.** On the pinned checkpoint at an 8 GiB budget, "The
+  capital of France is" completes to "**Paris**".
+
+## Not supported
+
+Text only. Not converted, not executed, and refused rather than half-done:
+
+* the **vision tower** (`model.visual.*`, 333 tensors) and everything that
+  configures it — no `vision.json` is written, so a container cannot be
+  handed an image;
+* **video and audio**;
+* the **MTP layer** (`mtp.*`, 31 tensors) and speculative decoding;
+* **native Qwen serving** — `serve/` has no Qwen chat format, so the
+  OpenAI-compatible server falls back to `chatfmt.py`, which refuses
+  tools, thinking and images by name rather than dropping them.
+
+The chat template is carried into the container as metadata, as every
+other release's is. Nothing reads it.
--- a/src/model.c
+++ b/src/model.c
@@ -1563,6 +1563,7 @@ static void cfg_from_json(waste_config *c, const js_doc *d, int cfg)
     c->index_dim   = (int)js_int(d, js_get(d, cfg, "index_head_dim"), 0);
     c->index_tail  = js_get(d, cfg, "index_kpool_always_select_tail") >= 0;
     c->tok_han_split = js_bool(d, js_get(d, cfg, "tokenizer_han_split"), 1);
+    c->tok_digit_run = (int)js_int(d, js_get(d, cfg, "tokenizer_digit_run"), 3);
 
     int lac = js_get(d, cfg, "linear_attn_config");
     c->full_rank_gate = js_get(d, lac, "use_full_rank_gate") >= 0;
--- a/src/model.c
+++ b/src/model.c
@@ -28,6 +28,11 @@
 #include "platform.h"
 #include "threads.h"
 #include "kda.h"
+#include "qwen_gdn.h"
+#include "qwen_hc.h"
+#include "qwen_moe.h"
+#include "qwen_ple.h"
+#include "qwen_qsa.h"
 #include "simd.h"
 #include "waste_backend.h"
 #include "waste_metal.h"
@@ -1115,7 +1120,8 @@ static int load_trunk(waste_model *m, const char *dir, const js_doc *d, int trun
              * row is read per token. Keeping 1.11 GB resident to touch 7 KB
              * of it is a bad trade against the expert cache, so leave it on
              * disk and pread the row. */
-            if (strstr(t->name, "embed_tokens.weight")) {
+            if (strstr(t->name, "embed_tokens.weight") ||
+                strstr(t->name, "ngram_head.")) {
                 t->on_disk = 1;
                 t->file_off = off;
                 t->file_scale_off = soff;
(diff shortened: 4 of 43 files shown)


mfethe1:
Independent CPU-only validation of `c8040c6686cf125c42ed585950cc38517f93d2de` against base `c66c7b3f`, on Apple M4 / darwin arm64:

- Project `make check`: base **60 passed / 0 failed / 14 skipped**; PR **77 passed / 0 failed / 15 skipped**. The additional skip is real-checkpoint Qwen tokenizer parity, not exercised here.
- Built-library `make serve-check` on the PR: **244 tests run, OK (3 skipped)**.
- Removing router renormalization (`w[j] /= s`) or the duplicate-selection guard (`used[best] = 1`) makes the Qwen numerical oracle checks fail. This is useful evidence that they discriminate numerical defects rather than merely mirror configuration constants.

These are synthetic-container/CPU results, not real-checkpoint tokenizer parity, model-quality validation, GPU verification, or an independent replication of the throughput figures in the PR. They support the tested paths without establishing full model compatibility.

For the next complementary validation, is there a preferred pinned checkpoint/tokenizer fixture and minimal acceptance command you would like independent testers to use? I would rather cover that remaining gap than repeat the synthetic tests.

Disclosure: this validation was performed by an AI-assisted research workflow; the scope above is deliberately limited to the recorded tests.

mfethe1:
Follow-up on my earlier review: I said the Qwen tokenizer parity check was skipped in my run. I closed that gap — **it passes against the real pinned checkpoint's tokenizer**, and it discriminates.

**Fixture, without the weights.** `docs/QWEN.md` points at the full release, but skip 77 only needs the tokenizer. Fetching three files from the pinned revision is enough:

```
Qwen/Qwen3.8-Flash-Next @ de4b8e4d43b917e7706784d8bb445c9af86a3540
  tokenizer.json         12,809,320 B  sha256 0997f410c57a1f4e53b09e4be8f4a172d90edd9564368fb0847030937229b9f3
  tokenizer_config.json      17,928 B  sha256 b11349aafa7cdc6a320767cf7ceb29ed82f7eda5d65e8e0819e76f0ce947bf27
  config.json                 4,745 B  sha256 889658f2508e8c61d409b02e70e0d78d8d4452ec65aaafbe129805d213d2e74b
```

`WASTE_QWEN_SRC=<that dir> python tests/test_qwen_tok.py` → **17 checks ok, exit 0** on `c8040c6686cf125c42ed585950cc38517f93d2de`. The digit-run cases are the ones that matter and they hold against the release's own tokenizer: `'2026'` → `[17, 15, 17, 21]` (four pieces, not one), `'It cost 1234567 yen in 2026.'` splits every digit, `'3.14159'` → seven ids, and the vocabulary genuinely contains no multi-digit token. Han: `'你好'` → `[109266]` with `han_split` false.

**Two negative controls, because a parity test that reads the same config it validates would pass either way.**

1. Made `KNOWN_PATTERNS` report `digit_run = 3` for every pattern (the Kimi/GLM value) → **FAILED, exit 1**.
2. Left the config alone and made the engine ignore it — `next_piece()` in `src/tokenizer.c` hardcoded `cnt < 3` instead of `cnt < digit_run` → **FAILED** on exactly the right assertion: `\p{N} keeps every digit apart: [256, 50, 54]`, i.e. `"2026"` collapsing to a merged piece.

So the test detects a mismatch on both sides of the boundary, not just a self-consistent restatement. Both mutations reverted; tree clean.

**Full suite with the fixture present:** `79 passed, 0 failed, 13 skipped` (was 77/0/15 for me without it). The second recovered skip was free — `chat.json` tool rendering vs K2's template needs only `chat_template.jinja` (2,021 B) from `moonshotai/Kimi-K2-Instruct`, not the checkpoint, via `K2_DIR`. Only `K3_DIR` (XTML vs `encoding_k3.py`) is still skipped here; I have no K3 release.

**Suggestion, entirely yours to take or drop:** the two fixture-gated skips resolve with ~12.8 MB of small files rather than a checkpoint. A note in `docs/QWEN.md` — or a `tools/fetch_tokenizer.sh` fetching just those three pinned files — would let reviewers and CI run the parity arm without the weights. Happy to send that as a patch if useful.

Scope, stated plainly: CPU tokenizer/pre-tokenizer parity only. No weights were downloaded, no generation was run, and nothing here speaks to numerical correctness of the kernels.

Environment: macOS arm64 (Apple M4), `make check` from a clean tree, `tokenizers` 0.23.2 in a throwaway venv.

marcobambini:
I cannot merge this as-is because I reproduced a heap-buffer-overflow on this PR's head (`c8040c6`) under AddressSanitizer. Generate a synthetic container with `tools/make_test_container.py --qwen`, change only `config.ple_conv_kernel_size` from `4` to `0`, then run `test_forward` with tokens `3,7,11,5`. The original container passes; the modified one produces an out-of-bounds read.

The loader accepts zero and allocates `ple_ring` with a minimal size, while the forward path falls back to a kernel size of four. Please reject invalid dimensions or normalize them consistently before validation, allocation, and execution, and add a regression test before merging.

mfethe1:
Confirming Marco's heap-buffer-overflow on `c8040c6` — reproduced independently under ASan+UBSan (Apple M4, clang), and the root cause is a disagreement inside the loader rather than a bad read at the use site.

**Reproduction.** Synthetic Qwen container, `ple_conv_kernel_size` `4` → `0`, everything else identical. Control loads and forwards; the zero container gives `heap-buffer-overflow READ` inside `qwen_dilated_conv_step` off `m->ple_ring`.

**Why.** Three places disagree about what a kernel of `0` means:

- the ring is allocated `(ple_conv_k - 1) * ngram_size` → **0 elements**;
- `REQUIRE_DATA` validates the conv weight with `c->ple_conv_k > 0 ? c->ple_conv_k : 4`, so it silently checks the weight *as if* the kernel were 4 and passes;
- the step then walks `KS = c->ple_conv_k > 0 ? c->ple_conv_k : 4` → **4** taps across a zero-length ring.

So the two defensive `> 0 ? … : 4` fallbacks are what cause this: they let a container the loader should have refused reach the step wearing a kernel it never declared. `cfg_sane()` validates `ngram_size`, `rotary_dim`, `hc_count` and friends — `ple_conv_k` was simply missing from that list.

**Fix I'd suggest** — refuse, don't normalize, matching how every other Qwen shape in `cfg_sane()` is handled:

```c
if (c->ple_conv_k < 1 || c->ple_conv_k > 64) return 0;
```

and then delete both `> 0 ? … : 4` fallbacks, so one stated meaning survives end to end. I deliberately avoided defaulting `0` → `4`: that would make the container's own declaration a lie and hide the next malformed one.

**Verified:**

- `make asan` (ASan + UBSan, `-fno-sanitize-recover=all`): **76 passed, 0 failed, 17 skipped** — up from 75, with a new `qwen_refused "a PLE conv kernel the ring cannot hold"` case beside the existing `layer_types` refusal tests.
- The zero container is now refused by both the loader and `waste info`, with no sanitizer report.
- The valid control is bit-for-bit unchanged: logits SHA-256 `bebc0f7d7580f0d3…` identical before and after.

Regression coverage is in `tests/run.sh` so it runs in the normal suite rather than as a one-off. Happy to send this as a PR to your branch if that's useful, @Skibisky — I'd rather not push a competing patch over the top of yours.

Skibisky:
Yes, that would be great. Thanks.

mfethe1:
Delivered as Skibisky/warp#1 against your branch — `ple_conv_k` bounded 1..64 in `cfg_sane()` for containers that declare a PLE layer (the field stays 0 and meaningless for kimi/GLM, where my first unscoped attempt refused their containers — caught by the suite, worth stating here), both `> 0 ? … : 4` fallbacks deleted, and a `qwen_refused` regression case in `tests/run.sh`. `make asan`: 76/0/17 vs base 75/0/17; valid-container logits bit-for-bit unchanged (sha256 `469a7433…` both arms). Happy to rebase if you land other changes first.

mfethe1:
Maintainer follow-up: the PLE ring-safety patch is now merged via Skibisky/warp#1 at `0d1366fff3e5ca9a651356e7b613e43fb14fb6ef`, which is also this PR’s current head. This closes the delivery loop on the zero-kernel overflow reported above; #63 itself remains open.

@marcobambini The change is available for re-review against your original malformed-container reproduction. It bounds the PLE kernel for PLE-bearing configurations and removes the inconsistent fallback interpretation, with a refusal regression in the normal test script. This is not a request to waive any other merge gates.

The earlier CPU sanitizer evidence remains narrowly software-safety evidence; this update adds confirmation of integration into the PR branch, not a new GPU, generation-quality, throughput, or current-head tokenizer-parity result.

AI-agent-assisted follow-up, posted with human authorization.

Skibisky:
I have been working on the Qwen3.8 implementation and got to a point where it gets ~12 tokens/sec on M4 at 'long contexts'. Initially it was ~6 tokens/sec, the work seems to have 'byte identical' output so I feel there is no downside in adding it to this PR. 

---

Since the initial PR, the Qwen path has received another performance pass, improving measured generation from roughly 6 tok/s to 12 tok/s at both short and long contexts. The main gains come from keeping the worker pool active across HyperConnection, parallelizing GDN and QSA work, reducing repeated quantization and dispatch, staging routed experts by residency, adding router lookahead, using i8mm for the trunk, and processing QSA scores and values in larger NEON-friendly batches.
The branch has also been merged with current upstream main (v0.8.0), including the new DeepSeek-V4.1 work. Conflicts were resolved additively so both architectures retain their tokenizer, converter, state, model, and test paths. The optimizations preserve output: the suite verifies byte-identical results across Qwen’s dispatch, cache, lookahead, prefill, and parallel execution variants, while the component and container-native oracle checks also pass. The complete post-merge suite reports 90 passed, 0 failed.

---

I have taken this opportunity to also merge latest in because the Deepseek addition touched the same files so now it should merge a little cleaner. I am not able to test with the Deepseek model so I am unsure if there are any regressions.

marcobambini:
Reviewed this on an Apple-silicon laptop with the K3 and Kimi-Linear containers on disk. Clean build, no warnings, and `tests/run.sh` reports **98 passed, 0 failed, 11 skipped** — all 18 Qwen checks green, including the container-native oracle, the bit-identical chunked prefill, the five refusals and the expert-schedule comparison. (GitHub Actions has not run yet; it needs approval for a fork PR.)

This is good work, and it is written *in* the repo's idiom rather than beside it: comments that name the failure that motivated them, refusals by name instead of silent drops, SKIP rather than a quiet pass, kernels isolated behind an independent PyTorch oracle, and a `docs/QWEN.md` that measures what it claims and has an explicit "Not supported" section. `waste.h` is untouched, `serve/` is untouched, the container stays format v0. The `raw_bytes()` rewrite in `convert.py` is a real improvement for every family, not just Qwen.

Three things I would like fixed before merge, two of which I verified rather than inferred, plus one design question that is the maintainer's call.

---

### 1. Heap overflow in a write, from an untrusted container (verified under ASan) — #69

`cfg_sane()` does not bound `shared_expert_intermediate_size`, but `m->ff` is sized from `max(dense_inter, moe_inter)` (`src/model.c:3272`) while `qwen_moe_layer`'s batched router projection writes `2 * shared_inter` into it (`src/model.c:8169-8172`).

A manifest declaring `shared_expert_intermediate_size: 512` over `moe_intermediate_size: 16`, with tensors shaped to match, passes `validate_qwen_tensors` and then:

```
ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 4 ... in waste_mvq4_rows_i8mm simd_i8mm.c:73
  #1 mvb_pieces model.c:1034
0 bytes after 128-byte region allocated at model.c:3272
```

Without ASan it runs to completion and prints logits. Either bound it in `cfg_sane` (`shared_inter <= max(moe_inter, dense_inter)`) or size `m->ff` from `shared_inter` too. It happens to fit exactly on Flash-Next — `shared_expert_intermediate_size` is 640 and so is `moe_intermediate_size` — which is why nothing noticed.

### 2. SIGFPE on x86 when `ngram_size` is 1 — #69

`src/model.c:1542`:

```c
const int width = (c->ple_embed && c->heads_per_ngram)
    ? c->ple_embed / ((c->ngram_size - 1) * c->heads_per_ngram) : 8;
```

`cfg_sane` admits `ngram_size == 1`, so the divisor is zero. On arm64 `sdiv` quietly yields 0 and the container is refused for a wrong shape; on x86 `idiv` raises `#DE`. I compiled the expression for both ISAs: `rc=0` on arm64, **`rc=136` (SIGFPE) on x86_64**. This is the same failure mode as the one already documented in this file at `src/model.c:3243` ("arm64's sdiv quietly yields 0, x86's idiv raises #DE ... issue #10"). Requiring `ngram_size >= 2` when `ple_layer >= 0` would cover it, and it is one more case in the existing `qwen_refused` block.

Smaller, same family: `linear_conv_kernel_dim` is not required to be >= 1 for Qwen — the `conv_k < 1` test in `cfg_sane` is on the non-Qwen branch — and with `KS == 0`, `gdn_conv_range` computes `ring + b * (KS - 1)`, a negative offset.

### 3. `docs/LEARNED.md` is not append-only on this branch

The new sections land as **§74–87** and the existing §74–79 are renumbered to **§88–93**, which also puts the dates out of order (§87 is 2026-09-16, §88 is 2026-09-15). That silently retargets six live cross-references:

| reference | meant | would mean |
|---|---|---|
| `CHANGELOG.md` §74 | a feasibility gate does not need the download | Qwen overview |
| `CHANGELOG.md` §75 | twenty-one strings is not a tokenizer corpus | where a Qwen decode step goes |
| `CHANGELOG.md` §76 | the oracle was wrong | one dispatch per Qwen layer |
| `CHANGELOG.md` §77 | a speculative batch of five | Qwen's trunk through i8mm |
| `CHANGELOG.md` §79 | three checks that were wrong | GDN's recurrence |
| `docs/DS41.md:592` §75 | the tokenizer corpus | where a Qwen decode step goes |

It looks like a rebase artifact — the branch predates 0.8.0 / DeepSeek-V4.1. Worth flagging that GitHub reports this as `MERGEABLE`, so it will not surface as a conflict: renumbering the new sections to §80–93 and restoring §74–79 in place is the fix.

---

### Design question, not a defect — #68

`src/model.c:2677` has a Qwen load call `waste_model_set_sdot4(TK_I8MM, ...)`, which writes the file-static `trunk_kern` that `matvec_t_inner` reads for *any* 4-bit tensor. On a machine with i8mm, a process that opens a Kimi container and then a Qwen one switches the K3 trunk to i8mm as well — arithmetic the comment itself says is not exact — including for a session already under way.

The comment and `docs/QWEN.md` both declare this, so it is a deliberate trade rather than an oversight. But it sits against two rules at once: `waste.h`'s "opaque `waste_ctx`, no global state", and the principle that results must not depend on incidental state — the same principle that makes `WASTE_XPAR` bit-identical by contract. A field on `waste_model` instead of a static would settle it. Maintainer's call.

### Nits

- `waste_qwen_ple_row_ids` zeroes `local_rows` before it checks `!local_rows`.
- `waste_qwen_qsa_select`: `nsel` is dead, hence the `(void)nsel`.
- `waste_qwen_hc_gates` recomputes the RMSNorm that `waste_qwen_hc_mix` already left in `scratch[0..H)`.
- `waste_model_get_info` returns `"qwen4_exp_text"` where the others return kebab names (`kimi-k3`, `glm5-next`); `tests/run.sh` had to add a case for it.
- `docs/FORMAT.md` does not document the new manifest keys (`layer_types`, `hc_count`, `indexer_*`, `ple_*`).
- The "bit-identical" claim for the four-wide QSA score path rests on the compiler contracting in one loop and not in the other. `tests/test_qsa_attn` checks it on the build machine, which is the right mitigation — just worth knowing it is what the guarantee rests on.
- No CHANGELOG entry or `WASTE_VERSION_*` bump, which is consistent with the rule as written (it is tied to a tag) — the tag that follows will need to carry all of this.

Short version: fix the two `cfg_sane` bounds with one more `qwen_refused` case each, rebase onto main with LEARNED.md renumbered at the end, and I am happy with this.

Filed the two that outlive this review: **#69** for the missing `cfg_sane` bounds (both of the above, plus `linear_conv_kernel_dim`), and **#68** for the process-wide trunk kernel.

Skibisky:
Changes have been made for 1, 2, and 3. 
Adjusted the name to keep things consistent.
Adding the manifest keys to the docs.

mfethe1:
I built both trees and ran a differential accept/reject matrix on `01e593c` against its base `8d5405d`. (`src/model.c` is byte-identical at your current head `68c4a40`, so all of this still applies after the naming commit.)

I used `tools/make_test_container.py --qwen` plus the `qwen_refused` harness (CPU only, no checkpoint downloaded). Five findings; the fourth is the one I think blocks the current bound.

**1. Two of the three new tests already pass on the unpatched tree.**

| mutation | base `8d5405d` | `01e593c` |
|---|---|---|
| `shared_expert_intermediate_size = max(moe,dense)+1` | REJECT | REJECT |
| `ngram_size = 1` | REJECT | REJECT |
| `linear_conv_kernel_dim = 0` | **ACCEPT** | **REJECT** |

Only the GDN-kernel case flips. The other two already fail on base, but for an unrelated reason — they die in the tensor checker, not the config checker:

```
waste: required tensor is missing or has the wrong shape: model.layers.0.mlp.shared_expert.gate_proj.weight
waste: required tensor is missing or has the wrong shape: model.layers.1.ple.ple_embedding.ngram_head.0.weight
```

Since `qwen_refused` asserts only on exit status, both pass today and would keep passing if the new bound were reverted. They do not pin the behaviour they describe.

**2. The shared-expert test cannot be repaired by editing the manifest alone.** The generator sizes the shared-expert tensors from `moe_intermediate_size`, never from `shared_expert_intermediate_size`:

```
t.quant(m + "shared_expert.gate_proj.weight", [moe, hid])
t.quant(m + "shared_expert.up_proj.weight",   [moe, hid])
t.quant(m + "shared_expert.down_proj.weight", [hid, moe])
```

So any manifest edit desynchronises the declared width from the emitted shapes and the shape check fires first, on every tree. To exercise the overflow the generator has to emit the shared-expert tensors at `shared_expert_intermediate_size`. I made that change in a scratch copy (three lines) and it does produce a container that is internally consistent *and* wide — happy to send it as a patch if useful.

**3. With that fixture, the defect you are fixing is real and reachable.** A tensor-consistent container with `shared_expert_intermediate_size = 32`, `moe_intermediate_size = 16`:

| | base `8d5405d` | `01e593c` |
|---|---|---|
| wide, tensor-consistent | **ACCEPT** | REJECT |

Base opens it, and `qwen_moe_layer` then writes `m->ff[0..shared_in)` and `m->ff[shared_in..2*shared_in)` into a buffer allocated at `2 * max(dense_inter, moe_inter)`. So the bug is worth guarding — my objection is only to where the guard is placed.

**4. The upper bound rejects a shipped Qwen checkpoint.** From the published `config.json` files:

| model | `intermediate_size` | `moe_intermediate_size` | `shared_expert_intermediate_size` | `max_inter` | passes `shared <= max_inter`? |
|---|---|---|---|---|---|
| Qwen1.5-MoE-A2.7B | 5632 | 1408 | 5632 | 5632 | yes |
| Qwen3-Next-80B-A3B-Instruct | 5120 | 512 | 512 | 5120 | yes |
| Qwen2-57B-A14B | 18944 | 2560 | **20480** | 18944 | **no** |

`shared_expert_intermediate_size` is an independent architectural dimension, not bounded by the other two, and Qwen2-57B-A14B ships with it larger than both. `if (c->shared_inter > max_inter) return 0;` refuses that container. I have not converted that checkpoint, so this is arithmetic against the guard text rather than an observed refusal of a real container — but the numbers are the model's own.

**5. The lower bound contradicts the forward path.** Two sites treat `0` as "fall back to the MoE width":

```c
const int shared    = c->shared_inter ? c->shared_inter : c->moe_inter;  /* qwen_shared_expert */
const int shared_in = c->shared_inter ? c->shared_inter : inter;         /* qwen_moe_layer */
```

`shared_inter < 1` refuses the value those lines are written to handle (base ACCEPT → `01e593c` REJECT for an explicit `0`). Omitting the key is fine either way, because the parser defaults it to `c->moe_inter`, so only an explicit `0` is affected. Either the sentinel should go, or the bound should be `< 0`.

**Suggestion, since the allocation is the thing that is actually too small.** Sizing the scratch from the width that indexes it removes both problems:

```c
int mx = c->dense_inter > c->moe_inter ? c->dense_inter : c->moe_inter;
if (c->shared_inter > mx) mx = c->shared_inter;
m->ff = (float *)calloc((size_t)2 * mx, sizeof(float));
```

with the config check relaxed to `if (c->shared_inter < 0) return 0;`. I built that variant and re-ran the matrix:

| case | base | `01e593c` | alloc-sized variant |
|---|---|---|---|
| pristine | ACCEPT | ACCEPT | ACCEPT |
| explicit `shared = 0` | ACCEPT | **REJECT** | ACCEPT |
| key omitted | ACCEPT | ACCEPT | ACCEPT |
| wide, tensor-consistent | ACCEPT | REJECT | ACCEPT |

Boundary on that last column: it is validated at container-open only. I did not run a forward pass, so it shows the geometry is accepted and the buffer is sized to cover the writes, not that decode output is correct.

**On the SIGFPE at the `width` division** — I could not reach it. `ngram_size = 1` with PLE enabled aborts earlier at `REQUIRE_MATRIX(... ple.key_proj.weight ...)`, and no run produced a signal. The generator does not emit the PLE tensor set, so I cannot confirm whether `ngram_size >= 2` actually covers the reported crash. If you have a container that reaches that line I will re-run against it.

The GDN-kernel guard is a clean catch and the only one of the three that changes behaviour today — worth keeping regardless of what happens to the shared-expert bound.

marcobambini:
Thanks both — I've pushed three commits on top of `68c4a40` (with `main` merged in) and will merge once CI is green. What they fix, and why:

### 1. The shared-expert bound did not close the overflow — `3b1c4f6`

`shared ≤ max(moe, dense)` keeps `m->ff` big enough, but `m->xq` — which quantizes the down projection's input for the i8mm trunk kernel Qwen defaults to — is sized from hidden, dense and the HyperConnection streams, and `moe` is none of them. A tensor-consistent container with `moe = shared = 2048` over a 32-wide hidden state passes the bound, opens, and on the first token:

```
WRITE of size 256 ... in quant_act4_mm_group model.c:667
0 bytes after 1280-byte region, allocated at waste_model_load model.c:3304
```

@mfethe1 — you were right that the bound was in the wrong place, and right about Qwen2-57B-A14B (I checked the config: shared 20480 against dense 18944). That bound was my suggestion in #69, so that one is on me. But sizing `m->ff` alone, as you proposed, leaves this exact write in place: same fixture, same trace. So the width is now bounded like its siblings (`1..2^20`) and every buffer a vector of that width goes through is sized from it — `m->ff`, `m->xq`, `m->xs` — with `waste_plan_memory` planning the same bytes. The `shared_inter ? : moe_inter` fallbacks are gone, per your point 5.

### 2. The tests could not see any of it — same commit

Your point 1 holds, and it is worse than a false pass: `qwen_refused` counted *any* non-zero exit, so a crash counted as a refusal too. It now requires `cfg_sane`'s message and an exit status that is not a signal. `make_test_container.py --qwen-shared N` generates a shared expert N wide in both tensors and manifest (your point 2), and `run.sh` opens and runs one 2048 wide and puts it through the container-native oracle — which reports the width it ran, so that arm cannot pass on the default fixture. Under `make asan` that forward pass is the overflow check. Each new check fails on a tree without its fix:

| | `8d5405d` | `68c4a40` | now |
|---|---|---|---|
| shared width 0 refused | accepted | PASS | PASS |
| `ngram_size 1` refused by `cfg_sane` | refused by a later check | PASS | PASS |
| conv kernel 0 refused by `cfg_sane` | accepted | PASS | PASS |
| shared 2048 opens and runs (ASan) | **overflow** | **refused** | PASS |

On the SIGFPE: it is reachable. With UBSan's `integer-divide-by-zero` on `8d5405d` and `ngram_size: 1`, the synthetic fixture stops at `src/model.c:1542:36: runtime error: division by zero`. The `ngram_head.0.weight` refusal in your matrix comes *after* that division — arm64 returns 0, which becomes a shape refusal; x86's `idiv` traps.

### 3. Fifteen LEARNED references pointed at DeepSeek entries — `acc98c2`

The renumbering moved the Qwen sections by six and updated the references between them, but the fifteen in `CLAUDE.md`, `docs/QWEN.md`, `src/model.c` and `src/qwen_qsa.c` kept the old numbers. The i8mm comment in `waste_model_load` pointed at "a speculative batch of five", for example. Each one is now shifted by six to where its section is.

### 4. `make asan` failed on arm64 — `84369e3`

`test_qsa_attn` passed at `-O2` and failed at `-O1` (`-0.0781467855 vs -0.0781467929`). The reference is the old `s += q * k`, and the compiler decides how that rounds: separately at `-O2`, as one fused multiply-add at `-O1`. The engine's scalar tail had the same split, so at `-O1` a selection's tail tokens were scored differently from the ones done four at a time. Both scalar score loops now keep the product and the sum as separate statements, which is what `-O2` already compiled them to — **release logits are byte-identical**. CI would not have caught this: its asan job runs on x86 without `-mfma`, where nothing can fuse.

**Results:** `tests/run.sh` 103 passed / 0 failed / 11 skipped; `make asan` 96 / 0 / 17; the default fixture and its logits byte-identical to before. #68 (the process-wide `trunk_kern`) is still open and tracked on its own; it does not block this.

marcobambini:
CI's first run on this branch was red, on three checks that had never run on it before. None of them came from the changes above, and all three reproduce on `68c4a40`. Two more commits:

### `test_qsa_attn` on linux-arm64 — `46721e9`

40 cases of 40 differed under gcc. I reproduced it in an `ubuntu:24.04` arm64 container, and isolated it: `-ffp-contract=off`, `on` and `fast` all fail, and **only turning the vectorizer off passes**. gcc turns the reference's one-accumulator dot product into in-order vector products, each rounded on its own, and leaves the kernel's four-accumulator loop scalar and fused. `84369e3`'s separate statements fixed clang at `-O1` and could not fix this: gcc contracts across statements, and vectorizes the two loops differently however they are written.

So the rounding is now written into the code rather than left to the compiler. `waste_qwen_qsa_mac` in `qwen_qsa.h` is `fmaf` wherever the target can fuse (`__ARM_FEATURE_FMA`, `__FMA__`) — one rounding by definition, one instruction — and two statements where it cannot, which is also where no compiler can fuse. The four-wide scores, the one-token scores, the scalar value tail and the test's reference all use it, and the NEON value path is only taken where it is fused as well. 0 cases of 40 differ on clang at `-O1` and `-O2`, and on gcc/arm64 at `-O1`, `-O2` and `-O3`, with contraction off, and under ASan.

**This changes numbers on arm64.** For a selection of 32 tokens or more — the four-wide path — scores now round once per element, where clang rounded twice. x86 without `-mfma` has no fused instruction to use, so it is unchanged. The synthetic fixture selects fewer than 32 tokens, so its logits stay byte-identical and cannot show the change; the container-native oracle agrees at 1.9e-06 on the default fixture and 2.4e-06 on the wide one. I don't have the real checkpoint here, so @Skibisky, if you can rerun `kernel_kl` on it, that would put a number on the change — I expect it to be noise, but I haven't measured it.

### Two Windows failures — `85d29e4`

The fixture check built its paths inside the `python -c` program text. MSYS2 rewrites POSIX paths in a native program's argv but not inside a `-c` string, so on Windows it opened `/tmp/...` and reported the fixture as not loading. The path is now passed as an argument, as the other checks already do. And `test_convert_qwen.py` printed `→`, which the Windows console code page can't encode, so it crashed in `print()`; it prints `->` now.

mfethe1:
I measured the `46721e9` rounding change on this merged tree (`09fcff3`, Apple clang 17.0.0, arm64-darwin24.6.0). Two things in the note above are narrower than what the code does, and the fixture turns out to be able to show the change after all.

### The change is not confined to selections of 32 or more

`46721e9` rewrote the four-wide score loop, but it rewrote the **one-token tail** at `qwen_qsa.c:245` the same way (`s += q*k` becomes `waste_qwen_qsa_mac`), and that loop is the whole of a sub-32 selection. So selections *below* 32 move too.

Isolating `46721e9` in-process against its own parent (same inputs, same order, one binary, and `v2` vs `v2` as a self-control that must be zero — it was, at both levels):

| n_sel | elements differing | max rel L2 |
|---|---|---|
| 4 | 21.94% | 8.21e-08 |
| 8 | 32.00% | 9.69e-08 |
| 16 | 42.31% | 1.19e-07 |
| 31 | 48.88% | 1.57e-07 |
| 32 | 52.88% | 1.34e-07 |
| 64 | 57.19% | 2.11e-07 |
| 2048 | 81.31% | 5.43e-07 |

(head geometry `D=16, Hq=4, Hkv=2`; identical row-for-row at `-O1` and `-O2`.) Nothing about the table changes at the 32 boundary, which is what you would expect once both loops call the same helper.

For contrast, `84369e3` alone — the separate-statements commit — *is* boundary-sensitive, and in the opposite direction to the four-wide framing: at `-O1` it changes numbers at 4, 8, 16, 24, 31 and 33, is exactly zero at 32, and zero at 64 and above; at `-O2` it is zero everywhere. That is `qwen_qsa.c:212`, `const int n4 = n_sel >= 32 ? n_sel : 0;` — at exactly 32 the four-wide path consumes the selection and leaves no tail, so the only loop that commit touched has no iterations.

### The synthetic fixture does select 32 or more, and does show the change

It selects fewer than 32 only at the default 16-token id list. I instrumented `waste_qwen_qsa_attn_heads` on the container fixture (`qwen4_exp_text`, 2 layers, `head_dim=16`, `indexer_budget=32`, layer 1 `full_attention`) and counted every call:

- T=16 (the default ids): 0 calls with `n_sel >= 32`
- T=64: 132 calls, max `n_sel` = 35
- T=256: 900 calls, max `n_sel` = 35

So the four-wide path runs, just not on 16 tokens. Extending the id list is enough.

Then, checksumming the attention output itself at both revisions (`WASTE_THREADS=1`, so the head workers cannot interleave; each binary run twice and byte-stable both times):

- attention output differs in **66.74%** of elements (10,935 of 16,384 over 1,024 calls)
- absolute difference: median 3.7e-09, p99 3.0e-08, **max 5.96e-08** (against median |out| 3.0e-02, max 5.2e-01)
- logits at T=64 and T=256: **byte-identical**

The mandatory control for a byte-identical result: the two `test_forward` binaries genuinely differ (`qwen_qsa.o` 8440 vs 7592 bytes, different md5s), and the value-path `#if` now requires `__ARM_FEATURE_FMA`, which holds here — so the same NEON block runs at both revisions and, at `head_dim=16`, the 16-wide loop consumes the whole head and the changed scalar remainder never executes. The score loops still differ, which is where the 5.96e-08 comes from.

So I would put it slightly differently than "its logits stay byte-identical and cannot show the change": with more than 16 tokens the fixture *does* exercise the change, and byte-identical logits under a measurably perturbed attention output is a real (if small) piece of evidence that the perturbation is absorbed before the logits on this model — 4-bit grouped activation quantization plus the bf16 K/V round-trip are both downstream of it.

### What I could not check

No gcc on this machine (`/usr/bin/gcc` is Apple clang), so I cannot reproduce the gcc/arm64 40-of-40 result or the vectorizer isolation — that part I am taking as read. And this is a 0.4 MB synthetic container, not a checkpoint: a 5.96e-08 attention perturbation being invisible at `vocab_size=256` says very little about a real vocabulary. @Skibisky, `kernel_kl` on the real checkpoint is still the number that matters here; I would also expect noise, but the fixture cannot stand in for it.

Skibisky:
Had a look at running kernal_kl on the real checkpoint. Twice to be sure. It seems to be within acceptable noise levels.

---

<html>
<body>

<p>I ran the long <code dir="ltr">kernel_kl</code> comparison on the <strong>real Qwen3.8-Flash-Next checkpoint</strong>: 5,918 teacher-forced prompt tokens followed by 256 generated positions. I adapted the harness to compare <code dir="ltr">46721e9^</code> with <code dir="ltr">46721e9</code> in one process, holding the i8mm trunk kernel fixed. Stock <code dir="ltr">kernel_kl</code> compares trunk kernels, so it cannot make this revision comparison directly.</p><div><div><div>

| Measure | Parent → 46721e9 |
| -- | -- |
Prompt mean KL | 0.00501
Prompt max KL | 0.189 at position 1323
Prompt argmax agreement | 5,759/5,918
Routed experts agreeing | 98.26%
Prompt perplexity | 4.731 → 4.732
Generated mean KL | 0.00192
Generated argmax agreement | 254/256

</div></div></div><p>A 128-position parent-versus-parent control was exactly zero. Mean KL was <strong>0.00634</strong> immediately before the 2,048-token boundary and <strong>0.00596</strong> immediately after it; the completed windows show no growing KL trend.</p><p>The rounding change has a measurable effect on real-checkpoint logits and routes, while prompt perplexity is effectively unchanged at the harness’s printed precision. </p>

</body>
</html>

---

<html>
<body>

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">I reran this on the real checkpoint. Setup: <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">qwen38-flash-next.waste</code> (124 GB, 48 layers, 512 experts top-10, <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">head_dim</code> 256, <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">indexer_budget</code> 2048, vocab 248,320), M4 Pro / 48 GiB, <strong>Apple clang 21.0.0</strong>, arm64-darwin25.6.0, 1,844 tokens of real text teacher-forced. I isolated 46721e9 against its own parent inside the merged tree — <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">src/qwen_qsa.{c,h}</code> are the only files differing, both sides built from <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">make clean</code>, <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">qwen_qsa.o</code> md5s differ. Controls first, because two of them nearly cost me the result: the same binary rerun is byte-identical, and <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">kernel_kl 0 vs 0</code> is exactly zero on all 1,844 positions and all 88,512 routings. One correction to the record while I was there — my first run showed <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">test_qsa_attn</code> failing 40 of 40 at HEAD, but that was a stale <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">qwen_qsa.o</code> in my own incremental revert, not a real failure. On clean builds it is <strong>0 of 40 at -O2 on clang 21</strong>, and so is the parent, which independently confirms your "-O2 is zero everywhere" result for 84369e3 on a compiler four major versions newer than the one you used. Worth noting <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">kernel_kl</code> can't actually answer this question by itself — it switches a <em>trunk</em> kernel inside one process — so I used it for the control and the yardstick, and did the cross-revision diff with a small dump harness built against each revision's <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">libwaste.a</code>.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Your narrower reading is the right one, and it goes further than the fixture could show. The first differing routing is at <strong>token 17, layer 33</strong>, where the selection is 17 — the one-token tail, not the four-wide path — so the change is not confined to selections of 32 or more. Before that flip the effect is pure arithmetic and small: positions 1–16 give KL ≤ 3.5e-06 and rel L2 1e-4 to 1e-3, with position 0 exactly zero. But <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">route_diff.py</code> calls that first disagreement <strong>a disagreement and not a tie</strong> — relative margin 7.378e-04, 74× the measured <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">--eps</code>. In context, that margin sits at the <strong>1st percentile</strong> of this model's own 3,072 boundary gaps (30 decisions were tighter and held; the minimum is 5.4e-05), so it is a tight call but not an unresolvable one — unlike the K3 tie in §71/<code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">route_diff</code>'s docstring, which was the single tightest decision in its run. Past token 17 the two runs are different trajectories, and the divergence accumulates with context: layer-routings differing grow from 3.2% in the first 256 tokens to 19.2% in the last, concentrated in layers 32–45, with layers 0–4 and 6 never differing at all.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">So: <strong>not byte-identical and not confined to ≥32, but well inside the band this engine already ships.</strong> The yardstick is <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">kernel_kl 0 vs 2</code> on the same prompt — f32 against i8mm, which is what a Qwen load selects when <code style="font-family: monospace; color: rgb(140, 140, 140); background: none 0% 0% / auto repeat scroll padding-box border-box rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">WASTE_TRUNK_KERNEL</code> is unset, i.e. the default's own deviation from exact. Your change is smaller on every comparable axis:</p>

|over 1,844 positions | 46721e9 vs parent | f32 vs i8mm (the shipped default)|
| -- | -- | -- |
argmax agreement | 1789/1844 = 97.02% | 1780/1844 = 96.53%
routed-expert agreement | 98.72% | 98.05%
layer-routings differing | 10495/88512 = 11.86% | 15512/88512 = 17.53%
KL, max | 2.05e-02 | 2.24e-01
KL, mean | 2.8e-03 (30 sampled positions) | 6.64e-03 (all positions)

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">And on the trade actually being decided — next-token NLL of the real text — neither side is better: HEAD wins 12 of 29 sampled positions and 29 of 63 in a finer run, loses 16 and 33, and the mean sign flips between the two samples (−0.32% and +0.84% ppl) against i8mm's +0.04% over the whole prompt. That is noise in the sense that matters: measurable, but a lateral move with no degradation, and roughly half the size of a difference already shipped by default on this model.</p>

</body>
</html>

mfethe1:
Thanks for running it twice on the real checkpoint — that closes the gate I pinned earlier, and the parent-vs-parent zero control plus the `qwen_qsa.o` md5 check are what make the rest readable. I re-derived the comparison arithmetic from the two runs you posted (no new compute on my side; everything below comes from numbers already in your comment) and I think your headline conclusion holds but the safety margin is roughly half of what the table implies.

**The two tables are not measured the same way.** The yardstick row (`kernel_kl 0 vs 2`, f32 vs i8mm) is a mean over *all* 1,844 positions, but the change's mean on that same line is over **30 sampled positions**. Your longer run does give a full-coverage prompt mean — 0.00501 over 5,918 positions. Comparing full coverage to full coverage:

| mean KL | value | ÷ yardstick mean 6.64e-03 |
| -- | -- | -- |
| as tabled (30 sampled, n=1,844) | 2.8e-03 | 0.42 |
| full coverage (n=5,918 run) | 5.01e-03 | **0.76** |

So "smaller on every comparable axis" survives, but on mean KL the margin is ~1.8× tighter than the sampled figure suggests. Routed-expert agreement moves the same direction: disagreement is 1.28% on the 1,844-token run (0.66× the yardstick's 1.95%) but 1.74% on the 5,918-token run (0.89×). That is consistent with your own observation that layer-routing divergence grows 3.2% → 19.2% across context — which means the ratio to the yardstick is *not* scale-free, and the 1,844-token comparison is the optimistic end of it.

**Max KL is the one I would not read at all right now.** The change reaches 0.189 on the long run versus a yardstick max of 0.224 — 0.84×, not the 0.09× the short-run table shows. But max is an extreme-value statistic and the long run has 3.2× the positions, so that ratio is confounded in the change's favour *and* against it depending on which side you fix. Neither 0.09× nor 0.84× is a valid statement; the yardstick max simply has not been measured at n=5,918.

**One axis does not narrow, and it is worth saying so:** argmax disagreement is 2.69% on the long run vs 2.98% on the short, against a 3.47% yardstick. Top-1 behaviour does not appear to degrade with context even though routing does.

**The NLL sign test supports you, and I can put a number on it.** 12 of 29 and 29 of 63 give two-sided sign-test p = 0.57 and p = 0.70 — nowhere near evidence of degradation. The honest caveat is width, not direction: the Wilson interval on 29/62 decisive positions is [0.35, 0.59], so the data are equally consistent with a true 59% loss rate. "No detectable degradation" is well supported; "no degradation" is not yet, at these sample sizes.

Open question, and it is a cheap one since your harness already does it: **what is `kernel_kl 0 vs 2` on the 5,918-token prompt?** A same-length yardstick would settle both the mean ratio and the max comparison in one run, and if the yardstick's divergence also accumulates with context then the ratio may well stay flat and your original framing gets stronger rather than weaker. As it stands the ≤0.42×/0.09× reading is measured at the short end only.

Boundary on my side: this is arithmetic over your published numbers plus standard sign-test/Wilson statistics. I have not run the checkpoint, and I am not claiming any independent measurement of this revision.

Issue #71, opened 2026-09-17
`convert.py` without `--reclaim` writes a DeepSeek-V4.1 container that can never be opened: `build_engram_meta()` is only called on the `--reclaim` path


helenkwok:
A DeepSeek-V4.1-Flash conversion that reports success produces a container which
every runtime subcommand rejects with `open: malformed container`. The container
is not damaged — the Engram *index* is simply never written, because
`build_engram_meta()` is called on one of the two trunk paths in `convert.py` and
not the other. The default path is the one missing it.

Cost of the failure: ~6.5 h of conversion and a 320 GB artifact that cannot be
loaded, with no diagnostic naming what is wrong.

## Symptom

```
$ waste info /ds41.waste        -> open: malformed container   (exit 1, stderr)
$ waste run  /ds41.waste "..."  -> open: malformed container
$ waste tokenize /ds41.waste x  -> open: malformed container
$ waste plan /ds41.waste        -> WORKS
      FLOOR                      4.86 GB
      one token's working set     2.97 GB
      fully resident           195.00 GB
```

`waste plan` parses the manifest and returns correct DS41 numbers, so the
manifest is fine. Only `waste_open()` fails.

## Why there is no diagnostic

`engram_open()` is the one `-2` in the load path that prints nothing —
`src/model.c:1964-1966`:

```c
snprintf(path, sizeof path, "%s/engram.json", dir);
char *es = slurp(path, NULL);
if (!es) return -2;          /* silent */
```

Every other `WASTE_E_FORMAT` site (`model.c:2160, 2186, 2194, 2218, 2231`,
`validate_text_tensors`/`bad_tensor`) `fprintf`s what failed. Since DS41 is the
only architecture with `engram_n != 0`, this is also the only arch that can fail
this way — and it fails with a bare five-word message.

## Root cause

`tools/convert.py` has two trunk paths. Only the first writes the Engram
metadata, and it writes it twice:

```python
# tools/convert.py:1619-1623  (the --reclaim path)
        if ds41:
            engram = build_engram(st, args.out, cfg, args.engram_bits)
            build_engram_meta(args.src, cfg, args.out)
            build_engram_meta(args.src, cfg, args.out)   # <- called twice
        reclaim(debt, args.reclaim, ShardDebt.TRUNK, "trunk")

# tools/convert.py:1990-1991  (the default path)
        if ds41:
            engram = build_engram(st, args.out, cfg, args.engram_bits)
            # <- build_engram_meta() is never called
```

[...]

PR #72, opened 2026-09-18
convert.py: write the Engram index on the default trunk path too

helenkwok:
Fixes the bug reported in #71.

`convert.py` has two trunk paths. Only the `--reclaim` one called
`build_engram_meta()`, and it called it twice; the default path called it not at
all. A DeepSeek-V4.1 conversion without `--reclaim` therefore writes the Engram
tables (110 GB here) and none of `engram.json`, and the container it reports as
finished cannot be opened — `engram_open()` needs that index.

Three changes:

- **`tools/convert.py`** — call `build_engram_meta()` on the default path, and
  drop the duplicate call on the `--reclaim` path. Net one line each way.
- **`src/model.c`** — `engram_open()` was the one `-2` in the load path that
  printed nothing, and DS41 is the only architecture that reaches it, so the
  failure was `open: malformed container` with no subject on a container whose
  files were all present and byte-exact. It now names the missing file.
- **`.github/workflows/ci.yml`** — a `guards` step asserting the invariant.

The guard walks the AST and checks **per block**, not by counting call sites.
That distinction is the point: with the duplicate in place the totals already
matched — two `build_engram()`, two `build_engram_meta()` — while the default
path wrote nothing. A counting guard passes on the broken tree; this one was
run against it and fails, naming `tools/convert.py:1984` and `:1991`.

No test without real weights can see either half of this, which is why the
invariant is checked on the source rather than in `tests/`.

### Verification

Full CI green on a fork, 11/11 jobs, including `build guards` with the new step
asserting at runtime (`ok: every block that writes Engram tables also writes the
index`):
https://github.com/helenkwok/warp/actions/runs/35290255815

Locally `make check` is 70 passed / 0 failed, unchanged from before the patch.

### Repair for containers already converted

No re-conversion needed if the source weights are still on disk (~3 min):

```python
import sys, json
sys.path.insert(0, "tools")
import ds41_engram
raw = json.load(open("SRC/config.json"))
cfg = raw.get("text_config") or raw     # build() does not unwrap text_config
ds41_engram.build("SRC", cfg, "ds41.waste")
```

Its own check passed here — the compressed token map came out 99092 ids,
matching the release's stated `engram_compressed_vocab_size`. `waste info` then
reports `552.37 B total, 16.62 B active/token` and the model generates.

Found converting `deepseek-ai/DeepSeek-V4.1-Flash` (475 GB staged, 320 GB
container) on an EPYC 7713.


marcobambini:
Thanks, @helenkwok. The diagnosis is right, and the per-block guard is the right design — for a reason that showed up while merging this. I've pushed a merge of `main` into your branch (`72b5b8a`) and will run CI on it.

**Why the merge was needed.** #63 landed after you branched, and it restructured the default trunk path. The `if ds41:` block you edited is now `if ds41 and not engram:`, one level out and after the Qwen PLE pass. It *still* called `build_engram()` without `build_engram_meta()`. So on `main` the bug was sitting somewhere your diff didn't touch, and git reported it as a conflict rather than applying your line there. Your guard caught it:

| tree | guard |
|---|---|
| current `main` | FAIL at `tools/convert.py:2445` |
| your branch, on its old base | ok |
| `main` merged, the call added to the new block | ok |

A counting guard would have passed a wrong resolution here, because the totals matched. So this is the case your design was written for. The merge takes `main`'s structure and adds the call to it. Your other hunk, dropping the duplicate on the `--reclaim` path, merged cleanly. Against `main` the PR is still exactly your three files, +50/−2.

Some suggestions, none of which block this:

- **`engram_open()` has more silent exits than the one you fixed.** On a synthetic DS41 container, a *corrupt* `engram.json` still fails as a bare `open: malformed container` — that is the `js_parse` check on the line right after yours. A missing `engram-tokmap.bin`, which the same `build_engram_meta()` writes, fails as `open: I/O error` with no file name. In all there are about a dozen `goto done` exits with nothing printed. A `why` string set at each one and printed once at `done:` would apply your principle to the whole function. Happy to take that as a follow-up PR.
- The new message says *"the Engram tables are there"*, but nothing has checked that when it prints. It also contains an em dash, which shows up as mojibake on a Windows console.
- **Repair.** With this fix in place, re-running the same `convert.py` command on a container that #71 broke should repair it: `build_engram` keeps tables already written at the right size (*"already written, keeping it"*) and then writes the index. I haven't run that, since it needs the weights. If it works on your EPYC box, it would be a simpler repair than the snippet.

PR #74, opened 2026-09-18
serve: model registry with hot-swap and per-request model logging

hendrikras:
Add a swappable model registry so the model behind a stable route name can change without reconfiguring clients.

- POST /v1/models/load swaps the loaded model; GET /v1/models lists the registry with loaded flags; generation requests are validated: unknown model is a 404, registered-but-not-loaded is a 409 pointing at the load endpoint instead of silently serving whatever is resident.
- CLI: --models PATH[=ID] registers swappable containers (repeatable), --keep-previous keeps a replaced model resident instead of unloading it.
- Swap opens the new container before closing the old one, so a failed load leaves the previous model serving; swap happens under the current engine's lock so a generation in flight finishes first.
- Request log lines now name the model: the one that served a success, the one refused on a 409/404, the one a load switched to. New --no-log-requests silences request logging.
- Tests over real sockets cover registry listing, swap, keep-previous, validation statuses, and the model names in the log; docs updated.


serve/server.py
```diff
@@ -317,27 +491,67 @@ def _health(self):
...
+            "previous": previous,
+            "models": [m for m in srv.registry if m in srv.engines],
+        })
 
     # ---- chat -----------------------------------------------------------
 
     def _chat(self):
         body = self._read_body()
         srv = self.server
         engine = srv.engine
```
marcobambini, line 547: `engine` is taken here, before any lock, and the swap only holds the
*previous* engine's lock. Two interleavings that break:

- Without `--keep-previous`: a request reads `srv.engine` (A) → a swap takes
  A's lock, opens B, closes A → the request takes A's lock and calls
  `state_reset()` / `generate()` on a closed engine. `_check()` raises
  instead of crashing, but the client gets a 500 for a request that arrived
  before the swap did.
- In both modes: everything else the request reads — `srv.chat_format`
  (L593), `srv.default_thinking` (L591), `srv.model_info` (L597),
  `srv.new_parser` (L603), `srv.stop_tokens` — comes from `srv` at the time
  it is read, i.e. possibly from the model swapped in after `engine` was
  taken. With `--keep-previous` that means building an XTML prompt for a
  container that speaks chat.json (or vice versa) and generating it on the
  other model's `waste_ctx`.

`_detect()` also rebinds those attributes one at a time, so even a request
that does not race the lock can read a half-updated set.

Suggestion: make the per-model state one immutable object (engine,
model_id, model_info, chat_format, stop_tokens, default_thinking, parser
factory, chat_error), have `_detect()` build a new one and assign it in a
single step, and have each request take a snapshot once, then lock
*that* snapshot's engine and re-check it is still current (or at least not
closed) after acquiring the lock — retrying on the new current model or
answering 409/503 if it moved.

serve/server.py
```diff
@@ -583,6 +797,9 @@ def _completions(self):
         """
         body = self._read_body()
         srv = self.server
+        self._log_model_from(body)      # before check_model_request: a 404 or
+        # 409 line should still name the model that was refused
+        srv.check_model_request(body)
```
marcobambini, line 802: (Anchored here because L812 is outside the diff: this is about `with srv.engine.lock:` at L812–835 in `_completions`.)

Same race, a bit sharper: `srv.engine` is read once to take the lock and
again on L813, L814 and L835. A request waiting on A's lock across a swap
wakes up holding A's lock and calls `state_reset()`, `tokenize()` and
`generate()` on B, while another request may be holding B's lock. Only
`generate()` takes the engine's RLock internally, so the reset/tokenize/
generate sequence of two requests can interleave on one `waste_ctx` — the
cross-request contamination the comment in `_chat` describes. Take the
engine into a local once, as `_chat` does, and apply the same snapshot
fix.

serve/server.py
```diff
@@ -153,6 +214,101 @@ def __init__(self, addr, handler, *, engine: Engine, model_id: str,
...
+        the server serving the model it was serving, and the only way to
+        guarantee that is not to have closed it yet. A failed swap costs
+        a load's worth of RAM for a moment; a swap that leaves no model
+        loaded costs the whole server. When two containers genuinely will
+        not fit together, size --budget so each open plans against what
+        the engine can actually get.
+
+        The old engine's lock is held across the whole swap, so a
+        generation in flight finishes before the slot moves under it, and
+        no request that took the old lock can find its engine closed.
```
marcobambini, line 268: "no request that took the old lock can find its engine closed" only holds
for a request that had *already acquired* the old lock. A request that
picked the old engine and is *waiting* for its lock gets it after the swap
has closed that engine (see L547). Either fix the race or correct the
docstring.

serve/__main__.py
```diff
@@ -217,7 +266,22 @@ def main(argv=None) -> int:
                     api_key=args.api_key,
                     default_max_tokens=args.max_tokens,
                     default_thinking=not args.no_thinking,
-                    allow_local_images=args.allow_local_images)
+                    allow_local_images=args.allow_local_images,
+                    log_requests=not args.no_log_requests,
+                    models=parse_registry(args.models),
+                    keep_previous=args.keep_previous,
+                    engine_kwargs={
+                        "ram_budget_bytes": args.budget,
```
marcobambini, line 274: With the default `--budget 0`, each `Engine()` sizes itself to ~3/4 of
`waste_usable_ram()`. Because the new container is opened before the old
one is closed, a swap has two contexts each planned against the whole
machine, ~1.5x usable RAM, and a fully-resident container also starts its
preload thread in that window. `docs/SERVE.md:353` says to "size
`--budget` so that moment fits", but nothing enforces it and the default
does not satisfy it.

Options, roughly in order of preference:
- Refuse `--models` without an explicit `--budget`, and check that
  2x budget fits under `waste_usable_ram()` (or plan the new container with
  `plan_memory` and refuse the swap with a 507/503 if it cannot fit next to
  the current one).
- With `--keep-previous` the sum is unbounded, not "both budgets": every
  model ever loaded stays open. Cap the number of resident engines, or
  count their budgets against a total and refuse the load that would
  exceed it.

serve/server.py
```diff
@@ -153,6 +214,101 @@ def __init__(self, addr, handler, *, engine: Engine, model_id: str,
                 # be asked to answer without it either.
                 self.default_thinking = fmt.think is not None
 
+    # ---- the model registry ----------------------------------------------
+
+    def check_model_request(self, body: dict) -> None:
```
marcobambini, line 219: This runs even when no `--models` were given, so a single-container server
now 404s any `model` other than its id. Its id defaults to the file name
(`k3.waste` → `k3`), and many OpenAI clients send a fixed name they cannot
easily change; the `serve --help` example itself sends `"model":"waste"`.
This is a breaking change for every existing deployment, not only for
those opting into the registry.

Suggestion: validate only when the registry has more than one entry, and
otherwise keep serving any name as today — or make strict validation its
own flag. Whichever way, it deserves a CHANGELOG line in the release that
ships it.

serve/__main__.py
```diff
@@ -140,8 +169,28 @@ def main(argv=None) -> int:
...
+                        "to the file name without .waste). Switching "
+                        "unloads the model it replaces unless "
+                        "--keep-previous")
+    s.add_argument("--keep-previous", action="store_true",
+                   help="keep a model resident when another is loaded. "
+                        "Off by default, and deliberately: the RAM two "
+                        "contexts need together is the sum of their "
+                        "budgets, and on the machines this engine targets "
+                        "that is the difference between working and "
+                        "paging. Both models can then answer at once — "
```
marcobambini, line 184: "Both models can then answer at once" — but `check_model_request` answers
409 for any model that is not the current one, resident or not, and
`load_model`'s docstring says so ("other resident models answer after the
next load names them, not before"). The help should describe what the
code does: switching back is a slot move instead of a reopen.

serve/server.py
```diff
@@ -317,27 +491,67 @@ def _health(self):
...
-                                      self.server.model_info)],
-        })
+        srv = self.server
+        # The whole registry, current model first, so a client scanning the
+        # list sees what is resident before what is only available. A
+        # registered-but-not-loaded entry carries no `waste` shape: its
+        # per-container facts are unknown until it is opened.
+        data = [api.model_object(srv.model_id, srv.started,
+                                 srv.model_info, loaded=True)]
+        data += [api.model_object(mid, srv.started, None, loaded=False)
```
marcobambini, line 501: With `--keep-previous`, a model that is still resident (in `srv.engines`)
is listed as `"loaded": false`, the same as one that was never opened. If
`loaded` means "current", call it that; if it means "resident", use
`mid in srv.engines`. The `/v1/models/load` response on L539 already
reports residency, so the two endpoints currently disagree.

serve/__main__.py
```diff
@@ -217,7 +266,22 @@ def main(argv=None) -> int:
...
+                    engine_kwargs={
+                        "ram_budget_bytes": args.budget,
+                        "ctx_tokens": args.ctx,
+                        "n_threads": args.threads,
+                        "cpu_list": args.cpus,
+                        "cache_policy": POLICIES[args.cache],
+                        "direct_io": not args.no_direct_io,
+                        "vision": args.vision,
+                        "verify_records": args.verify,
+                        "usage_path": args.usage,
```
marcobambini, line 282: `--usage` is a learned hotlist of one container's experts (default
`<model>/usage.waste`). When given explicitly, it is passed to every
container a swap opens, so GLM would be opened with K3's hotlist. With the
default (None) each container finds its own and this is fine. Either drop
`usage_path` from `engine_kwargs`, or refuse `--usage` together with
`--models`.

serve/server.py
```diff
@@ -65,18 +78,66 @@ def __init__(self, addr, handler, *, engine: Engine, model_id: str,
...
+    #
+    # The block that follows used to run once, in __init__, and every
+    # handler read its verdicts as constants for the life of the process.
+    # With a swappable registry they are per-model: a container without
+    # XTML markers must not inherit the previous container's chat format,
+    # stop tokens, or thinking default. So the same block is a method, run
+    # again on every load, and it keeps the old comment because every word
+    # of it still holds.
+
+    def _detect(self, engine: Engine, model_id: str) -> None:
```
marcobambini, line 130: At startup, `__main__` reports what the container can do: the banner, the
`chat_error` warning ("/v1/chat/completions is unavailable for this
container"), the thinking/tools/images line. A swap re-runs detection but
reports none of it, so switching to a container without a usable chat
format produces 400s with no operator-visible note. At least log a line on
stderr per swap: which model, and `chat_error` if set.

marcobambini (changes requested):
Thanks — this is a useful feature and it is carefully done: the swap-order
rollback, the real-socket tests and the docs are all in the right shape.
A few things need to change before it can go in, mostly around what happens
to requests that overlap a swap and around the RAM ceiling, plus one
behavior change that reaches servers that never use `--models`.

1. **Requests racing a swap** (server.py `_chat` / `_completions`): the
   engine is picked before its lock is taken, and the per-model state
   (`chat_format`, `stop_tokens`, `default_thinking`, `model_info`) is read
   from `srv` rather than from a snapshot, so a request that overlaps a swap
   can run on a closed engine, or on the new engine holding only the old
   engine's lock, or on one model's `waste_ctx` with the other model's
   format. Inline comments below.
2. **RAM budget**: opening the new container before closing the old one
   with the startup `engine_kwargs` means that with the default
   `--budget 0` both engines auto-size to ~3/4 of usable RAM, so a swap
   briefly asks for ~1.5x the machine. `ram_budget_bytes` is meant to be a
   hard ceiling for everything the process holds, and paging is the failure
   mode this engine exists to avoid.
3. **The 404 on an unknown `model` also applies without `--models`**:
   servers that never opt into the registry start rejecting clients that
   send a fixed model name, including the `"model":"waste"` example in
   `serve --help`.

Smaller: `--keep-previous`'s help text promises concurrent answering that
the code does not do; an explicit `--usage` hotlist is handed to every
container; a swap skips the startup diagnostics (`chat_error` in
particular).

mfethe1:
Nice PR — the registry/404/409 semantics, the open-before-close rollback ordering, and the budget pre-check are all well shaped, and `check_engine` is exactly the right answer to the queued-request problem.

One thing I think should block merge: a lock-order inversion that wedges the whole server.

**`load_model` takes `_slot_lock` then the engine lock; the generation path takes them in the opposite order.**

- `load_model` — `with self._slot_lock:` (`serve/server.py:559`) → `with previous_engine.lock:` (`:572`)
- generation — `with engine.lock:` (`:882`, `:1121`) → `srv.check_engine(slot)` → `current_slot()` → `with self._slot_lock:` (`:322`)

A `POST /v1/models/load` arriving while a generation is in flight closes the cycle: the load holds `_slot_lock` waiting on the engine lock the generation holds, and the generation's `check_engine` waits on `_slot_lock`. Since every request path reaches `_slot_lock`, the server wedges for all clients, not just the two involved.

**Repro** — deterministic, no timing race, on a clean clone of this PR head (`1d2e287`), stdlib only, using the repo's own `FakeEngine`:

Start a thread that holds `engine.lock` (as a generation does), call `load_model("swap-a")` from another and give it time to reach `with previous_engine.lock:`, then have the first thread call the real `current_slot()`:

```
ARM 1 control  (no load in flight): returned in 0.0000s
ARM 2 test     (load in flight)   : HUNG >5s
```

**Suggested fix:** make the swap use the same order as everything else — read the slot, release `_slot_lock`, take `previous_engine.lock`, then re-take `_slot_lock` and re-check the slot did not move (retry if another swap won the race). With that applied both arms return in 0.0000s.

**Regression check on the fix**, `python3 -m unittest discover -s tests/serve -t .` (macOS arm64, py3.14.7):

| | result |
|---|---|
| this PR head, unpatched | Ran 394 tests — failures=2, skipped=83 |
| with the fix | Ran 394 tests — failures=2, skipped=83 |

Identical, so the reorder costs nothing in the suite.

---

Separately, those **2 failures are pre-existing on this branch** and look like a real second defect introduced by `1d2e287`:

```
FAIL: TestGlmToolsFromChatJson.test_load_log_names_the_model_loaded                   404 != 200
FAIL: TestGlmToolsFromChatJson.test_load_log_stderr_reports_swap_and_chat_error_if_set 404 != 200
```

Both `POST /v1/models/load {"model": "swap-a"}`, both get 404. That class's `setUp` (`tests/serve/test_server.py:1773`) never passes `models=` to `serve()`, so the registry is empty and `swap-a` is a correct 404 — the two new tests were added to a fixture that has no registry. Adding `"models": {"swap-a": ...}` to its `server_kwargs` should be all it needs.

Also FYI GitHub currently reports this unmergeable — content conflicts with main in `CHANGELOG.md` and `serve/server.py`.

Happy to send the lock-order fix as a PR against this branch if that's useful.

mfethe1:
Heads-up: the merge of `main` at 4d167855 accidentally broke the serve integration suite on this branch — it is not a flake, and it is not from the model-registry code.

**What happened.** The conflict resolution in `serve/server.py` inserted a duplicate copy of the pre-#73 `ChatServer` class *inside* `ModelSlot.detect`'s body: the file now has two top-level `class ChatServer` definitions, and the first 31 lines of the duplicate (starting `class ChatServer(ThreadingHTTPServer):`, ending `self.tmpdir = self._tmp`) sit between `default_thinking = start_thinking` and the `try:` that used to follow it. At module scope that terminates `detect` early, so `ModelSlot.detect` returns `None`, `ChatServer.__init__` stores `self._slot = None`, and every request path (`current_slot()`) raises `AttributeError: 'NoneType' object has no attribute 'engine'` — the handler thread dies and the client sees a dropped connection.

**Attribution** (macOS arm64, suite run at each of the three revisions in a clean clone):

| revision | tests/serve/test_integration.py |
|---|---|
| `main` @ 09fcff3 | 12/12 OK |
| branch pre-merge head `1d2e287` | 12/12 OK |
| merged head `4d167855` | 1 failure + 10 errors — all `TestRealStack` requests `RemoteDisconnected`, server log shows the `AttributeError` above at `serve/server.py` `_chat` |

CI agrees: on 4d167855 the `macos-arm64` job fails exactly there (`FAIL serve suite`, eight `TestRealStack` errors). It only shows up on runners that build the library — where `libwaste` is absent the class skips, which is why the windows `run` check still passed.

**Fix.** Deleting the duplicated 31-line block (keep main's `#73` `ChatServer` with `request_queue_size = 128` as the only definition, and restore `detect`'s original body from the branch side) takes the merged head back to 12/12 OK on the integration suite, and the full `tests/serve` tree back to exactly the two pre-existing `TestGlmToolsFromChatJson` 404 failures — which I confirmed also fail at `1d2e287`, so the merge introduces no other serve regression.

Repro, no build needed:

```
python3 - <<'EOF'
from serve.server import ModelSlot
import inspect
src = inspect.getsource(ModelSlot.detect)
print("detect body length:", len(src.splitlines()))   # ~85 when intact; ~15 when truncated
EOF
```

(and `grep -c '^class ChatServer' serve/server.py` → `2` on the current head, `1` everywhere else.)

mfethe1:
551a72f9 fixed the `ModelSlot.detect` breakage — the duplicated `ChatServer` block is gone and `request_queue_size = 128` is back where it belongs. Confirmed against the diff `4d167855..551a72f9` (1 file, +7/-31).

CI is still red on `linux-x86_64`, `linux-arm64`, `macos-arm64` and `windows-x86_64 (native build + suite)`, but **the merge is exonerated for what remains**. The `macos-arm64` job on 551a72f9 fails on exactly two tests and nothing else:

```
FAIL: tests.serve.test_server.TestGlmToolsFromChatJson.test_load_log_names_the_model_loaded
FAIL: tests.serve.test_server.TestGlmToolsFromChatJson.test_load_log_stderr_reports_swap_and_chat_error_if_set
AssertionError: 404 != 200
```

**Attribution** — copies of `def test_load_log_names_the_model_loaded` in `tests/serve/test_server.py`:

| ref | copies |
|---|---|
| `b34cb47d` strict model validation | 1 |
| `1364101f` loaded means residency | 1 |
| `e0b73381` refusing --usage with --models | 1 |
| `1d2e2873` serve: Logs a line on stderr per swap | 2 |
| `4d167855` merge of main | 2 |
| `551a72f9` current head | 2 |

The second copy arrived with `1d2e2873`, three commits before the merge.

**Root cause.** The block belongs to `TestRequestLogs` (line 918), which starts the server with `--models` so `swap-a` is in the registry. A second copy was also appended to the end of `TestGlmToolsFromChatJson` (line 1870), whose `setUp` builds a GLM `chat.json` container with no registry at all. `POST /v1/models/load {"model": "swap-a"}` is correctly a **404** there, so the copied `assertEqual(status, 200)` cannot pass. The trailing `def test_get_log_carries_no_model(self): pass` in the same block also silently overrides that case for the GLM class.

The same test names pass in their real home:

```
$ python -m unittest -v tests.serve.test_server.TestRequestLogs
test_load_log_names_the_model_loaded ... ok
test_get_log_carries_no_model ... ok
Ran 6 tests in 3.056s — OK
```

**Fix** — delete the 14-line block at the end of `TestGlmToolsFromChatJson` (1870–1883), keeping the `TestRequestLogs` copy.

**Measured** on a clean clone of `pull/74/head` @ `551a72f9`, libwaste rebuilt from that tree (`make libwaste.a libwaste.dylib libwastevq.dylib`), same machine and same binary for both arms:

```
before: Ran 394 tests — FAILED (failures=2, skipped=26)
after:  Ran 391 tests — OK (skipped=26)          # 1 file changed, 14 deletions(-)
```

The two failures in the before arm are exactly the two the CI job reports, and the after arm is clean with no remaining failures. (My earlier pass of this reused a prebuilt `libwaste.dylib` from another checkout, which added a spurious `test_version_matches_the_header` `0.8.0 != 0.8.1` failure; rebuilding removed it, so it was mine, not yours.)

If you do want the swap-logging assertions covered for the GLM container too, that class's `setUp` needs a registry that actually holds `swap-a` — but `TestRequestLogs` already covers the behaviour, so deleting is the straightforward call.

*Boundary: local macOS arm64, CPU-only, self-built libwaste, no GPU and no real checkpoint. These are pure Python `tests/serve` cases, which is why they reproduce identically on all four red runners.*

